PR #7, opened 2018-02-21
[WIP] surface-layers redesign

emersion:
**Status**: protocol is mostly finalized and only undergoing minor revisions at the moment. There are client and server implementations that have proven everything except for popups, which are still under development/review.

Surface layers is a protocol extension designed to make portable desktop shell components possible. Among the use-cases supported by this protocol are:

* "Background" surfaces like static or animated wallpapers or perhaps some kind of conky deal
* Drawer-style terminal emulators like that old KDE thing
* Panels (taskbars) complete with "start" menus, context menus, etc
* Launchers like dmenu and rofi
* Notification daemons like dunst or Ubuntu's thing
* Lock screens
* On screen keyboards
* Desktop shortcuts
* Tools that select a region of the screen for screenshots
* Android-style notification drawer
* Showing visual feedback on touch-screen touches
* Displaying keypresses visually for presentations, recordings, live streams, or whatever
* Adding decorations to a mouse to make it stick out more during presentations, recordings, etc
* And more...

All of these components could be used on any compositor that supports surface layers.

Note that access to this protocol should be limited to trusted clients. This is entirely feasible but out of scope for this discussion.


unstable/surface-layers.xml
```diff
@@ -115,12 +137,40 @@
...
     </request>
 
-    <enum name="input_devices">
+    <event name="configure">
+      <description summary="suggest a surface change">
+        The configure event asks the client to resize its surface.
+
+        The size is a hint, in the sense that the client is free to ignore it if
+        it doesn't resize, pick a smaller size (to satisfy aspect ratio or
+        resize in steps of NxM pixels).
```
SirCmpwn: What do we do if the client is anchored on two opposite sides of the screen but chooses to use a size smaller than that? We should define that behavior.

Ongy:
Would it be a good idea to document a few default namespaces and how we would order them?

As in: "virtual input" > "locking" > "overlays"
Where '>' is above.

emersion:
>Would it be a good idea to document a few default namespaces and how we would order them?

I don't really want to put these in the protocol. I'd rather add another file with a list of standardized namespaces (IANA-like).

Ongy:
How exactly does libwayland validate enums?

I would like to have my catch-all smantics on the exclusive input.
Can we just set the entire uint32 to `~0` for this, or would that break in some protocol validation?

emersion:
Since it's defined as a bitfield in the protocol, the value will be valid if it represents a (possibly empty) set of enum items. So I think a bit outside the enum is an error. Why can't you just set all items in the enum to 1?

Ongy:
because then my locking application would not lock a device it wasn't built for, because it did not exist yet.

If there is every another input capability on seat (BCI), every lock would have to be adjusted to explicitly lock that down as well.
And would have to check which ones are supported by the compositor, to not get the error on an older compositor implementation that does not support this device category yet.

If we get the capabilities in events and can just fill a local bitfield up during initialisation, that should work as well.
Though it should also work in a way that can catch hotplugging new devices while locked and other funny cases.

Ongy:
When we talked about layerin before, we had a usecase of a windows-esque bar, with a start menu.

For the current exclusive output zones of this to affect the main layouting area, the bar has to be bellow the main area (we only get exclusive upwards), but the start menu should be above the main windows.

We could do this, if we (optionally) allowed popups (or the positioners?) to be in a different layer than the main surface.

SirCmpwn:
>For the current exclusive output zones of this to affect the main layouting area, the bar has to be bellow the main area (we only get exclusive upwards), but the start menu should be above the main windows.

I addressed this earlier. Popups should be shown at `max(parent surface layer + 1, shell surfaces layer + 1)`.

unstable/surface-layers.xml
```diff
@@ -42,14 +41,38 @@
...
         layer_surface, or raises a protocol error if another role is already
         assigned.
+
+        Creating a layer surface from a wl_surface which has a buffer attached
+        or committed is a client error, and any attempts by a client to attach
+        or manipulate a buffer prior to the first layer_surface.configure call
+        must also be treated as errors.
+
+        Clients can specify a namespace that defines the purpose of the layer
+        surface.
```
SirCmpwn: Elaborate

unstable/surface-layers.xml
```diff
@@ -81,9 +239,11 @@
         depths. The compositor may choose under which circumstances to send
         inputs. By convention, the compositor will send input events to
         the top-most surface of the top-most layer.
+
+        Interactivity is double-buffered, see wl_surface.commit.
       </description>
-      <arg name="input_types" type="uint" summary="mask of input devices to use"/>
-      <arg name="exclusive_types" type="uint" summary="mask of input devices to use exclusively"/>
+      <arg name="input_types" type="uint" enum="wl_seat::capability" summary="mask of input devices to use"/>
+      <arg name="exclusive_types" type="uint" enum="wl_seat::capability" summary="mask of input devices to use exclusively"/>
```
Ongy: https://github.com/wayland-project/wayland-protocols/blob/master/unstable/tablet/tablet-unstable-v2.xml It looks like new input types won't be added to the `wl_seat::capability`, so this approach likely won't work to catch all input devices in the future

emersion: Not sure how we should handle input types from other protocols in this protocol. Providing our own enum isn't really a solution.

SirCmpwn:
I have pushed an updated protocol and am seeking buy-in. It uses xdg-shell popups and positioners instead of duplicating that functionality here.

unstable/sway-surface-layers-unstable-v1.xml
```diff
@@ -0,0 +1,232 @@
...
+        This assigns an xdg_shell::xdg_popup's parent to this layer_surface.
+        This popup should have been created via
+        xdg_shell::xdg_surface.get_popup with the parent set to NULL, and this
+        request must be invoked before committing the popup's initial state.
+
+        See the documentation of xdg_popup for more details about what an
+        xdg_popup is and how it is used.
+      </description>
+      <arg name="popup" type="object" interface="xdg_shell::xdg_popup"/>
+      <arg name="positioner" type="object" interface="xdg_shell::xdg_positioner"/>
```
emersion: The positioner is already provided to `xdg_surface::get_popup`

SirCmpwn: derp

unstable/sway-surface-layers-unstable-v1.xml
```diff
@@ -0,0 +1,232 @@
...
+      </description>
+      <arg name="horizontal" type="int"/>
+      <arg name="vertical" type="int"/>
+    </request>
+
+    <request name="get_popup">
+      <description summary="assign this layer_surface as an xdg_popup parent">
+        This assigns an xdg_shell::xdg_popup's parent to this layer_surface.
+        This popup should have been created via
+        xdg_shell::xdg_surface.get_popup with the parent set to NULL, and this
```
emersion: Should be `xdg_surface::get_popup`. Protocol names are just used to group interfaces, they aren't a namespace.

SirCmpwn: k

unstable/wlr-surface-layers-unstable-v1.xml
```diff
@@ -0,0 +1,231 @@
...
+      <description summary="assign this layer_surface as an xdg_popup parent">
+        This assigns an xdg_shell::xdg_popup's parent to this layer_surface.
+        This popup should have been created via xdg_surface::get_popup with the
+        parent set to NULL, and this request must be invoked before committing
+        the popup's initial state.
+
+        See the documentation of xdg_popup for more details about what an
+        xdg_popup is and how it is used.
+      </description>
+      <arg name="popup" type="object" interface="xdg_shell::xdg_popup"/>
```
emersion: Should be `xdg_popup` only, no need for that `xdg_shell` (will help the automatic documentation generator).

unstable/wlr-surface-layers-unstable-v1.xml
```diff
@@ -0,0 +1,231 @@
...
+    SOFTWARE, INCLUDING ALL IMPLIED WARRANTIES OF MERCHANTABILITY AND
+    FITNESS, IN NO EVENT SHALL THE COPYRIGHT HOLDERS BE LIABLE FOR ANY
+    SPECIAL, INDIRECT OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN
+    AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION,
+    ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF
+    THIS SOFTWARE.
+  </copyright>
+
+  <interface name="zwlr_surface_layers_v1" version="1">
```
acrisci: I've brought this up before but we never came to a satisfying answer.

When I first saw this name of this interface, I was really confused as to what it does. This interface (`zwlr_surface_layers_v1`) is a shell. All the other shells have "shell" somewhere in their name and I think there is a good reason to continue this.

Here are my suggestions:

* zwlr_layer_surface_shell_v1
* zwlr_surface_layers_shell_v1

SirCmpwn: Sounds reasonable to me

emersion: I disagree with `zwlr_layer_surface_shell_v1`. layer-surface isn't a shell.

SirCmpwn: The layer surface isn't a shell, but the thing creating layer surfaces is.

unstable/wlr-surface-layers-unstable-v1.xml
```diff
@@ -0,0 +1,231 @@
...
+    </enum>
+
+    <enum name="layer">
+      <description summary="available layers for surfaces">
+        These values indicate which layers a surface can be rendered in. They
+        are ordered by z depth, bottom-most first. Traditional shell surfaces
+        will typically be rendered between the bottom and top layers.
+        Fullscreen shell surfaces are typically rendered at the top layer.
+        Multiple surfaces can share a single layer, and ordering within a
+        single layer is undefined.
```
acrisci: I don't understand the difference between background/bottom and top/overlay. If I'm making a background or overlay application, it's easy to make the right choice, but what about other application types like bars and notifications?

What is the use case for picking "bottom" instead of "background"?

What is the use case for picking "top" instead of "overlay"?

SirCmpwn: background: wallpaper

bottom: panel (windows can be dragged over it)

top: notifications

overlay: lockscreen

acrisci:
There's a big oversight here in that there is no specification for the output that the surface layer wants to be on.

For a background application, each background layer surface should be different for every output. Duplicating the surface on all outputs will not be acceptable. Same for bar applications (which might have a configured "primary output").

How does the layer surface specify what output it wants to be on?

Do we want to allow a single layer surface to span multiple outputs?

SirCmpwn:
Derp, that is a big oversight. We do not want a single layer to span multiple outputs. I'll add an output to get_layer_surface.

SirCmpwn:
Wait, it's already there.

acrisci:
Oh ok I did overlook that. I think it wasn't in the older version of the protocol and I just assumed it wasn't there in this one.

unstable/wlr-surface-layers-unstable-v1.xml
agx: An on screen keyboard would need to go atop of the lock screen (since it might be needed to enter the password) 
Is this worth special casing?

unstable/wlr-surface-layers-unstable-v1.xml
```diff
@@ -0,0 +1,231 @@
...
+      <arg name="left" type="int"/>
+    </request>
+
+    <request name="get_popup">
+      <description summary="assign this layer_surface as an xdg_popup parent">
+        This assigns an xdg_popup's parent to this layer_surface.  This popup
+        should have been created via xdg_surface::get_popup with the parent set
+        to NULL, and this request must be invoked before committing the popup's
+        initial state.
+
```
agx: https://github.com/wayland-project/wayland-protocols/blob/master/stable/xdg-shell/xdg-shell.xml#L1010
mandates that the parent  of a xdg_popup needs to be a xdg_toplevel or xdg_popup surface so strictly speaking using a layer_surface would be a violation of the xdg protocol. I wonder if it might confuser toolkits somehow if this is violated.

unstable/wlr-surface-layers-unstable-v1.xml
emersion: This is fixed by the namespace: both surfaces will be in the overlay layer, one will have the `lockscreen` namespace, the other the `osk` namespace. The compositor will be able to put the OSK on top of the lockscreen.

We're still exploring other options to solve this issue though. Let us know if you have ideas!

unstable/wlr-surface-layers-unstable-v1.xml
emersion: You're right. I think the NULL parent in `get_popup` was originally designed for xdg-foreign.

To fix this, we could duplicate `xdg_popup` into `wlr_layer_popup` and keep using `xdg_positioner`.

SirCmpwn: >To fix this, we could duplicate xdg_popup into wlr_layer_popup and keep using xdg_positioner.

This is probably appropriate. But we should float a discussion on the mailing list for removing this restriction from xdg-shell, it wouldn't be a breaking change.

SirCmpwn:
I am of the opinion that we should move forward utilizing xdg popup instead of copy-pasting it into our protocol. Because making the change to xdg-shell would be backwards- and forwards-compatible and we have a compelling case for it, this approach is low-risk. But even if we can't get buy-in on making the change to xdg-shell, because the surface-layers protocol is unstable anyway we can always change it later. In that case we'll probably want to use xdg popups internally even if we don't commit that to the protocol, so refactoring our code with that in mind will be worthwhile anyway.

SirCmpwn:
Pulling Mir into the discussion: @AlanGriffiths @wmww 

We might want to loop in KDE at some point but I don't think they're terribly excited about surface-layers.

Ongy:
`seat::capability` may not be extended for new input devices re-using it will limit us to the devices that are added to the seat, which is a rather annoying and hard to predict restriction.

If the plan is to have a locking application use only `surface-layers` I would still prefer to have a way to catch all input devices, even those that the client doesn't know about because it has an older version of the protocol.

emersion:
>In that case we'll probably want to use xdg popups internally even if we don't commit that to the protocol, so refactoring our code with that in mind will be worthwhile anyway.

We won't be able to do so because xdg popups are created from xdg surfaces. If we copy-paste xdg popups in our protocol, they won't inherit from xdg surface (doing so will end up making the implementation very complicated) and they'll have their own configure/ack_configure events. Also xdg popups only have 2 requests and 2 events, it's not like it's a big interface.

>Input semantics are defined through a second, optional interface

I don't understand what's the value of this.

>Input semantics are defined per-seat

So if the client doesn't care about the seat, it will need to create one input object per seat anyway? This adds a lot of complexity and I don't find any use-case for only listening to one seat input events.

>The description gets into the nitty gritty details of how different kinds of events are handled etc

That's better.

Should the input state really be double-buffered?

>seat::capability may not be extended for new input devices re-using it will limit us to the devices that are added to the seat, which is a rather annoying and hard to predict restriction.

How is it "hard to predict"? I think it's totally fine for now, supporting tablet devices in surface-layers is not something we should focus on right now.

SirCmpwn:
>seat::capability may not be extended for new input devices re-using it will limit us to the devices that are added to the seat, which is a rather annoying and hard to predict restriction.

I think we can set "all" as `~1` and make setting unused bits acceptable.

>>Input semantics are defined through a second, optional interface
>
>I don't understand what's the value of this.

It makes it easier to do it per-seat.

>So if the client doesn't care about the seat, it will need to create one input object per seat anyway? This adds a lot of complexity and I don't find any use-case for only listening to one seat input events.

The client has to care about seats. Outside of surface layers you have to fuck with seats to make your clients work, too. Here's a use-case: dmenu. I don't see why dmenu would have to take over all of the seats instead of just the one that launched it.

>Should the input state really be double-buffered?

If it's separated into two requests, yes.

1. Type "he" into gedit
2. Layer surface calls set_events
3. Type "llo"
4. Layer surface calls set_passthrough
5. Type " world"

You have now typed "he world" into gedit.

I think separating them into two requests makes it cleaner anyway, if only because it makes the docs easier to read and understand.

>How is it "hard to predict"? I think it's totally fine for now, supporting tablet devices in surface-layers is not something we should focus on right now.

ongy has a good point that a lock screen would want to eat all input events, period.

SirCmpwn:
>We won't be able to do so because xdg popups are created from xdg surfaces. If we copy-paste xdg popups in our protocol, they won't inherit from xdg surface (doing so will end up making the implementation very complicated) and they'll have their own configure/ack_configure events. Also xdg popups only have 2 requests and 2 events, it's not like it's a big interface.

xdg popups will still be created from xdg surfaces in layer surfaces. The difference is that their parent will not be an xdg surface. I am strongly in favor of reusing xdg popups here instead of copying and pasting them in.

emersion:
>Here's a use-case: dmenu. I don't see why dmenu would have to take over all of the seats instead of just the one that launched it.

This is a focus issue. This has nothing to do with the client being able to select the seat.

>If it's separated into two requests, yes.

I was speaking of the double-buffered state, not the fact that it's two requests.

>xdg popups will still be created from xdg surfaces. The difference is that their parent will not be an xdg surface. I am strongly in favor of reusing xdg popups here instead of copying and pasting them in.

I was speaking of using xdg popups internally while exposing our own layer popups.

unstable/wlr-surface-layers-unstable-v1.xml
wmww: bottom: panel (windows can be dragged over it)
In some shells (Gnome comes to mind) panels are rendered on top of windows. Would it be appropriate for panels to request top in that case? Might it be a good idea to have something in between, so that shell components can let the shell decide?

SirCmpwn:
>This is a focus issue. This has nothing to do with the client being able to select the seat.

It does, because layer surface clients dictate how the seat handles focus.

>I was speaking of the double-buffered state, not the fact that it's two requests.

So was I. The two are linked: we need double buffering *because* we have two requests.

>I was speaking of using xdg popups internally while exposing our own layer popups.

We don't have layer popups. We just have xdg popups.

1. surface = xdg_shell.get_xdg_surface(surface=wl_surface)
2. positioner = xdg_shell.get_positioner()
3. popup = surface.get_popup(parent=NULL, positioner=positioner)
4. layer surface.get_popup(popup=popup)

It might be clearer if `layer_surface::get_popup` was instead called `assign_popup` or something.

acrisci:
> I pushed another attempt at defining input semantics with this protocol. A few compelling use-cases for passthrough make it difficult to justify not handling this in the protocol:

I'm strongly against fallthrough. This will require a complete redesign of the seat to handle more than one window with pointer focus for this use case or library users to define their own parallel seat that can handle this case.

The bottom three use cases can be handled much more simply by clients and the compositor with an interface to listen to global input events.

This is the remaining use case:

> Unity-style notifications, which fade out when the pointer moves over them and pass input through to the surfaces beneath

Is it really worth the effort to basically rewrite the seat (or tell people to write their own seat) for this?

SirCmpwn:
>I'm strongly against fallthrough. This will require a complete redesign of the seat to handle more than one window with pointer focus for this use case or library users to define their own parallel seat that can handle this case.

No, it won't. The changes will not be nearly as severe as you're making them out to be. I have trotted out use-case after use-case for designing the seat to accomodate for this and time and time again you've pushed back because it's hard. Fuck that. We're in the business of making shit work for our users.

>because of security concerns

Surface layers is already a security concern. We've known from day 1 that this is a protocol that you have to secure access to. If we move global input watching into a separate protocol then that's the same exact problem in a different form factor.

These uses cases **exist** and *must* be accomodated for.

SirCmpwn:
If you (@emersion or @acrisci) don't like this approach then you need to come forward with an equally capable approach.

unstable/wlr-surface-layers-unstable-v1.xml
SirCmpwn: Yes, panels that render on top of other windows should choose the top layer. I don't think we need something in-between.

acrisci:
> Surface layers is already a security concern.

The security concern without fallthrough is that a rogue overlay window could eat all the input events and stop the computer from being usable.

The concern with fallthrough is that any application you trust with surface layers can silently monitor your keystrokes without you knowing it.

These are a different class of security problems that require different thinking about who you trust to run applications on your system.

With a separate interface for global input handling, the user may be more liberal in what new bars they try because they can apply a lower standard of trust to surface layers applications. This will be beneficial to get users to adopt the protocol.

> I have trotted out use-case after use-case for designing the seat to accomodate for this

I've only heard the one use case and I don't think it is very compelling. I think you just have a feeling that you will think up some use for this in the future and you don't want to be limited now. You can convince me it is valuable by giving an example of a good feature that requires this.

> you've pushed back because it's hard

I'm pushing back against multiple windows with pointer focus because it doesn't fit the wayland protocol well. For instance, this breaks `wl_pointer::set_cursor` because now two or more windows will fight over it.

> If you (@emersion or @acrisci) don't like this approach then you need to come forward with an equally capable approach.

Can we just put it off and decide how to do it later? I'm sure we can think of something better than this.

SirCmpwn:
>Can we just put it off and decide how to do it later? I'm sure we can think of something better than this.

No, I consider solving this problem a basic requirement of surface layers.

SirCmpwn:
>The security concern without fallthrough is that a rogue overlay window could eat all the input events and stop the computer from being usable.

Don't forget phishing and annoyances (like blacking out the screen).

>The concern with fallthrough is that any application you trust with surface layers can silently monitor your keystrokes without you knowing it.

We need this feature somewhere. It's going to have to be provided by a protocol somewhere. There are real problems with not associating it with a surface, relative pointer events are important and there's no way - by design - for a layer surface to know it's actual coordinates on screen. Nor are screens guaranteed to be two dimensional or surfaces guaranteed to be arranged relative to each other in predictable two dimentional space either! What about a compositor for which, if there are multiple panels, only shows one at a time and makes you cycle between them? The semantics for this are not solved with a generic standalone "eat all input" protocol.

>I've only heard the one use case and I don't think it is very compelling. I think you just have a feeling that you will think up some use for this in the future and you don't want to be limited now. You can convince me it is valuable by giving an example of a good feature that requires this.

I've given four use-cases *in this thread*, and offhand I know that at least two of them have *existing software for X11*, one of which I know offhand has a similar application in Android, and the last of which I didn't know off of the top of my head but found software for X11 from a quick Google search. The use cases exist, god dammit.

>I'm pushing back against multiple windows with pointer focus because it doesn't fit the wayland protocol well. For instance, this breaks `wl_pointer::set_cursor` because now two or more windows will fight over it.

I wish you would have said this earlier because it's the only valid technical concern I've heard from you yet. This is easily solved, though, we'll just define it as using the cursor from the client being passed through to.

>Can we just put it off and decide how to do it later? I'm sure we can think of something better than this.

No, I consider solving this problem a basic requirement of surface layers.

---

I am open to the idea of doing input handling in a separate protocol, perhaps in a way that is useful across more shells and can be secured separately. But separating it won't be trivial. We can't set a blanket interactive or not interactive for a layer surface. Some surfaces will listen to touch and pointer events but not keyboard events, like an OSK, and shouldn't get keyboard focus. Some will want keyboard but not pointer or touch events, like rofi or dmenu. Some will want everything, like screen lockers. And that's even before considering passthrough.

I have had it up to here with putting out protocols and hearing bitching and moaning in return but no actual alternatives.

## These use cases exist and we will support them.

This is not negotiable. We are going to design support for this right now, not later. I will not entertain any more complaints about my approach to input unless there is an XML file attached with a suggested alternative. One of the *core values* of this project is solving these problems.

SirCmpwn:
Current version of this patch does not include passthrough. After further discussion on IRC it introduces a lot of problems and will drag out this process. We can do it separately later.

Current version is the same as before, but has passthrough removed. Gonna start working on support for this in wlroots.

unstable/wlr-layer-shell-unstable-v1.xml
```diff
@@ -0,0 +1,238 @@
...
+      Clients can use this interface to assign the surface_layer role to
+      wl_surfaces. Such surfaces are assigned to a "layer" of the output and
+      rendered with a defined z-depth respective to each other. They may also be
+      anchored to the edges and corners of a screen and specify input handling
+      semantics. This interface should be suitable for the implementation of
+      many desktop shell components, and a broad number of other applications
+      that interact with the desktop.
+    </description>
+
+    <request name="get_layer_surface">
```
agx: Wouldn't set_layer_surface be better (I want to become this surface a layer surface)?
Weston's desktop-shell protocol calls similar methods "set_panel", "set_backround" with e.g. this signature:

    <request name="set_panel">
      <arg name="output" type="object" interface="wl_outp

ut"/>
      <arg name="surface" type="object" interface="wl_surface"/>
    </request>

and it always sounded natural to me since I set the surface role.

SirCmpwn: No - this is a function on the layer surface shell, not on wl_surface, and it follows the conventions set by other shells. It gets an instance of the layer_surface type.

emersion: xdg-shell's wording is using "get" too, so it's probably better to stick with that (xdg-shell is official, weston's internal protocol isn't)

unstable/wlr-layer-shell-unstable-v1.xml
```diff
@@ -221,11 +252,20 @@
       <arg name="height" type="uint"/>
     </event>
 
+    <event name="closed">
+      <description summary="surface should be closed">
+    The closed event is sent by the compositor when the surface will no
```
emersion: The indentation is a little weird here

unstable/wlr-layer-shell-unstable-v1.xml
```diff
@@ -111,9 +127,24 @@
         implementation-dependent - do not assume that this region will not
         actually be occluded.
 
-        This value is only meaningful if the surface is anchored to an edge,
-        rather than a corner. The zone is the number of pixels from the edge
-        that are considered exclusive.
+        A positive value is only meaningful if the surface is anchored to an
```
emersion: Strictly positive

unstable/wlr-layer-shell-unstable-v1.xml
```diff
@@ -86,11 +86,27 @@
...
-      double-buffered, and will be applied at the time wl_surface.commit of the
-      corresponding wl_surface is called.
+      Layer surface state (size, anchor, exclusive zone, margin, interactivity)
+      is double-buffered, and will be applied at the time wl_surface.commit of
+      the corresponding wl_surface is called.
     </description>
 
+    <request name="set_size">
+      <description summary="sets the size of the surface">
+        Sets the size of the surface in pixels. The compositor will display the
```
emersion: Pixels is a little ambiguous here: is it in surface-local coordinates or in buffer-local coordinates? If the surface has a scale factor of 2 and a buffer size of 42, the surface size is 21.

So we should probably s/in pixels/in surface-local coordinates/.

carnager:
Any news on the KDE front? Imho wayland needs solutions that work everywhere. Ideally both KDE and gnome people agree to this and then one can make it a core protocol instead of yet another niche-solution.

emersion:
@carnager https://drewdevault.com/2018/04/28/KDE-Sprint-retrospective.html

PR #9, opened 2018-04-03
Add wlr-input-inhibitor-unstable-v1

SirCmpwn:


Timidger:
In there a reason "wlr" is included in the names of this protocol? Is there something specific in it I'm not seeing that requires it to be used with wlroots?

SirCmpwn:
Because we're the organization sponsoring the protocol

Timidger:
Then why is there not the same thing for surface layers?

SirCmpwn:
There is?

SirCmpwn:
https://github.com/swaywm/wlr-protocols/pull/7/files#diff-8d1419da79a9ab9a0cf9ab943a8d1b8f

Timidger:
oh in the PR. Ignore me I was looking at the latest in master

agx:
Nice!

Might make sense to disable gestures built into the compositor as well
```
 this resource will receive all input events normally. The compositor will
also disable all of its own keyboard shortcuts and touch gestures while the inhibitor is
active.
```

Also some surface-layer clients (e.g. OSK) might need to be allowed to unlock the screen.

SirCmpwn:
Yeah, I should clarify that some. Also should clarify that text_input clients may still receive enough input events to do their job.

SirCmpwn:
wlroots implementation: https://github.com/swaywm/wlroots/pull/807

unstable/wlr-input-inhibitor-unstable-v1.xml
```diff
@@ -0,0 +1,67 @@
...
+  <interface name="zwlr_input_inhibitor_v1" version="1">
+    <description summary="inhibits input to other clients">
+      While this resource exists, input to clients other than the owner of the
+      inhibitor resource will not receive input events. The client that owns
+      this resource will receive all input events normally. The compositor will
+      also disable all of its own input processing (such as keyboard shortcuts)
+      while the inhibitor is active.
+
+      The compositor may continue to send input events to selected clients,
+      such as an on-screen keyboard (via the input-method protocol).
```
acrisci: I think you could qualify this a bit better.

unstable/wlr-input-inhibitor-unstable-v1.xml
```diff
@@ -0,0 +1,67 @@
...
+    ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF
+    THIS SOFTWARE.
+  </copyright>
+
+  <interface name="zwlr_input_inhibit_manager_v1" version="1">
+    <description summary="inhibits input events to other clients">
+      Clients can use this interface to prevent input events from being sent to
+      any surfaces but its own, which is useful for example in lock screen
+      software. It is assumed that access to this interface will be locked down
+      to whitelisted clients by the compositor.
```
acrisci: I think you should specify what happens to input focus of the existing clients when this gets created (and destroyed).

unstable/wlr-input-inhibitor-unstable-v1.xml
SirCmpwn: It's deliberately vague

PR #11, opened 2018-04-22
A DMA-BUF export protocol

atomnuker:
This is just to track any issues rather than going over IRC.

### DIFF
--- a/unstable/wlr-export-dmabuf-unstable-v1.xml
+++ b/unstable/wlr-export-dmabuf-unstable-v1.xml
@@ -1,0 +1,203 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<protocol name="wlr_export_dmabuf_unstable_v1">
+  <copyright>
+    Copyright © 2018 Rostislav Pehlivanov
+
+    Permission is hereby granted, free of charge, to any person obtaining a
+    copy of this software and associated documentation files (the "Software"),
+    to deal in the Software without restriction, including without limitation
+    the rights to use, copy, modify, merge, publish, distribute, sublicense,
+    and/or sell copies of the Software, and to permit persons to whom the
+    Software is furnished to do so, subject to the following conditions:
+
+    The above copyright notice and this permission notice (including the next
+    paragraph) shall be included in all copies or substantial portions of the
+    Software.
+
+    THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+    IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+    FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
+    THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+    LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+    FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+    DEALINGS IN THE SOFTWARE.
+  </copyright>
+
+  <description summary="a protocol for low overhead screen content capturing">
+    An interface to capture surfaces in an efficient way by exporting DMA-BUFs.
+
+    Warning! The protocol described in this file is experimental and
+    backward incompatible changes may be made. Backward compatible changes
+    may be added together with the corresponding interface version bump.
+    Backward incompatible changes are done by bumping the version number in
+    the protocol and interface names and resetting the interface version.
+    Once the protocol is to be declared stable, the 'z' prefix and the
+    version number in the protocol and interface names are removed and the
+    interface version number is reset.
+  </description>
+
+  <interface name="zwlr_export_dmabuf_manager_v1" version="1">
+    <description summary="manager to inform clients and begin capturing">
+      This object is a manager with which to start capturing from sources.
+    </description>
+
+    <request name="capture_output">
+      <description summary="capture a frame from an output">
+        Capture the next frame of a an entire output.
+      </description>
+      <arg name="frame" type="new_id" interface="zwlr_export_dmabuf_frame_v1"/>
+      <arg name="overlay_cursor" type="int"
+           summary="include custom client hardware cursor on top of the frame"/>
+      <arg name="output" type="object" interface="wl_output"/>
+    </request>
+
+    <request name="destroy" type="destructor">
+      <description summary="destroy the manager">
+        All objects created by the manager will still remain valid, until their
+        appropriate destroy request has been called.
+      </description>
+    </request>
+  </interface>
+
+  <interface name="zwlr_export_dmabuf_frame_v1" version="1">
+    <description summary="a DMA-BUF frame">
+      This object represents a single DMA-BUF frame.
+
+      If the capture is successful, the compositor will first send a "frame"
+      event, followed by one or several "object". When the frame is available
+      for readout, the "ready" event is sent.
+
+      If the capture failed, the "cancel" event is sent. This can happen anytime
+      before the "ready" event.
+
+      Once either a "ready" or a "cancel" event is received, the client should
+      destroy the frame. Once an "object" event is received, the client is
+      responsible for closing the associated file descriptor.
+
+      All frames are read-only and may not be written into or altered.
+    </description>
+
+    <enum name="flags">
+      <description summary="frame flags">
+        Special flags that should be respected by the client.
+      </description>
+      <entry name="transient" value="0x1"
+             summary="clients should copy frame before processing"/>
+    </enum>
+
+    <event name="frame">
+      <description summary="a frame description">
+        Main event supplying the client with information about the frame. If the
+        capture didn't fail, this event is always emitted first before any other
+        events.
+
+        This event is followed by a number of "object" as specified by the
+        "num_objects" argument.
+      </description>
+      <arg name="width" type="uint"
+           summary="frame width in pixels"/>
+      <arg name="height" type="uint"
+           summary="frame height in pixels"/>
+      <arg name="offset_x" type="uint"
+           summary="crop offset for the x axis"/>
+      <arg name="offset_y" type="uint"
+           summary="crop offset for the y axis"/>
+      <arg name="buffer_flags" type="uint"
+           summary="flags which indicate properties (invert, interlacing),
+                    has the same values as zwp_linux_buffer_params_v1:flags"/>
+      <arg name="flags" type="uint" enum="flags"
+           summary="indicates special frame features"/>
+      <arg name="format" type="uint"
+           summary="format of the frame (DRM_FORMAT_*)"/>
+      <arg name="mod_high" type="uint"
+           summary="drm format modifier, high"/>
+      <arg name="mod_low" type="uint"
+           summary="drm format modifier, low"/>
+      <arg name="num_objects" type="uint"
+           summary="indicates how many objects (FDs) the frame has (max 4)"/>
+    </event>
+
+    <event name="object">
+      <description summary="an object description">
+        Event which serves to supply the client with the file descriptors
+        containing the data for each object.
+
+        After receiving this event, the client must always close the file
+        descriptor as soon as they're done with it and even if the frame fails.
+      </description>
+      <arg name="index" type="uint"
+           summary="index of the current object"/>
+      <arg name="fd" type="fd"
+           summary="fd of the current object"/>
+      <arg name="size" type="uint"
+           summary="size in bytes for the current object"/>
+      <arg name="offset" type="uint"
+           summary="starting point for the data in the object's fd"/>
+      <arg name="stride" type="uint"
+           summary="line size in bytes"/>
+      <arg name="plane_index" type="uint"
+           summary="index of the the plane the data in the object applies to"/>
+    </event>
+
+    <event name="ready">
+      <description summary="indicates frame is available for reading">
+        This event is sent as soon as the frame is presented, indicating it is
+        available for reading. This event includes the time at which
+        presentation happened at.
+
+        The timestamp is expressed as tv_sec_hi, tv_sec_lo, tv_nsec triples,
+        each component being an unsigned 32-bit value. Whole seconds are in
+        tv_sec which is a 64-bit value combined from tv_sec_hi and tv_sec_lo,
+        and the additional fractional part in tv_nsec as nanoseconds. Hence,
+        for valid timestamps tv_nsec must be in [0, 999999999]. The seconds part
+        may have an arbitrary offset at start.
+
+        After receiving this event, the client should destroy this object.
+      </description>
+      <arg name="tv_sec_hi" type="uint"
+           summary="high 32 bits of the seconds part of the timestamp"/>
+      <arg name="tv_sec_lo" type="uint"
+           summary="low 32 bits of the seconds part of the timestamp"/>
+      <arg name="tv_nsec" type="uint"
+           summary="nanoseconds part of the timestamp"/>
+    </event>
+
+    <enum name="cancel_reason">
+      <description summary="cancel reason">
+        Indicates reason for cancelling the frame.
+      </description>
+      <entry name="temporary" value="0"
+             summary="temporary error, source will produce more frames"/>
+      <entry name="permanent" value="1"
+             summary="fatal error, source will not produce frames"/>
+      <entry name="resizing" value="2"
+             summary="temporary error, source will produce more frames"/>
+    </enum>
+
+    <event name="cancel">
+      <description summary="indicates the frame is no longer valid">
+        If the capture failed or if the frame is no longer valid after the
+        "frame" event has been emitted, this event will be used to inform the
+        client to scrap the frame.
+
+        If the failure is temporary, the client may capture again the same
+        source. If the failure is permanent, any further attempts to capture the
+        same source will fail again.
+
+        After receiving this event, the client should destroy this object.
+      </description>
+      <arg name="reason" type="uint" enum="cancel_reason"
+           summary="indicates a reason for cancelling this frame capture"/>
+    </event>
+
+    <request name="destroy" type="destructor">
+      <description summary="delete this object, used or not">
+        Unreferences the frame. This request must be called as soon as its no
+        longer used.
+
+        It can be called at any time by the client. The client will still have
+        to close any FDs it has been given.
+      </description>
+    </request>
+  </interface>
+</protocol>


emersion:
The way Wayland solves this multiple buffer types issue is by creating a `wl_buffer` which can be either backed by `wl_shm` or by DMA-BUF. Maybe we could re-use `wl_buffer` here too, by introducing another object for DMA-BUF exports (one similar to `zwp_linux_buffer_params_v1`, but s/request/event/). This would allow:
* To simplify this protocol: it would only have to deal with layers, not with DMA-BUF planes
* To be able to use other buffer formats in the future (this has proven to be useful with e.g. `wl_drm` vs. DMA-BUF)

SirCmpwn:
You should run this through wayland-scanner and fix the errors.

atomnuker:
Having a higher level abstraction (like wl_buffer) wouldn't simplify anything, it would complicate matters. This is about as low level as we can get down to. I don't see an issue with exposing multiple dmabuf objects, they always go around together.
I don't see how what's described can prevent future buffer formats from being implemented. Just add a new type and specify what the file descriptors will contain.

Also I get no wayland-scanner errors.

unstable/zwlr-screencontent-v1.xml
```diff
@@ -0,0 +1,209 @@
...
+    <event name="receive_layer">
+      <description summary="layer receiving callback">
+        Callback which serves to supply the client with information on what's
+        contained in each file descriptor and how its laid out.
+        Will be called after the main receive event, once per layer.
+      </description>
+      <arg name="index" type="uint"
+           summary="index of the current plane"/>
+      <arg name="format" type="uint"
+           summary="format of the layer (DRM_FORMAT_*)"/>
```
emersion: There's an enum for this

unstable/zwlr-screencontent-v1.xml
```diff
@@ -0,0 +1,209 @@
...
+        Main callback supplying the client with information about the frame,
+        as well as an object to serve as context for destruction. Always called
+        first before any other events.
+      </description>
+      <arg name="width" type="uint"
+           summary="frame width, scaling factor included"/>
+      <arg name="height" type="uint"
+           summary="frame height, scaling factor included"/>
+      <arg name="scale" type="uint"
+           summary="current scaling factor of the client"/>
```
emersion: There's a `scale`, but why isn't there a `transform`?

unstable/zwlr-screencontent-v1.xml
```diff
@@ -0,0 +1,209 @@
...
+      <arg name="type" type="uint" enum="frame_type"
+           summary="indicates the type of frame preferred"/>
+      <arg name="frame_flags" type="uint" enum="frame_flags"
+           summary="indicates special frame features"/>
+      <arg name="mod_high" type="uint"
+           summary="drm format modifier, high"/>
+      <arg name="mod_low" type="uint"
+           summary="drm format modifier, low"/>
```
emersion: This is DMA-BUF specific right?

unstable/zwlr-screencontent-v1.xml
```diff
@@ -0,0 +1,209 @@
...
+        first before any other events.
+      </description>
+      <arg name="width" type="uint"
+           summary="frame width, scaling factor included"/>
+      <arg name="height" type="uint"
+           summary="frame height, scaling factor included"/>
+      <arg name="scale" type="uint"
+           summary="current scaling factor of the client"/>
+      <arg name="type" type="uint" enum="frame_type"
+           summary="indicates the type of frame preferred"/>
```
emersion: "preferred" should be removed here

unstable/zwlr-screencontent-v1.xml
ascent12: Yes.

emersion: This is one of the reasons why I don't like having one type for all buffer types. We'll just end up putting here all properties for all buffer types and say "this one is only used if the type is XXX".

unstable/zwlr-screencontent-v1.xml
```diff
@@ -0,0 +1,209 @@
...
+      <arg name="num_objects" type="uint"
+           summary="indicates how many objects (FDs) the frame has (max 4)"/>
+      <arg name="num_layers" type="uint"
+           summary="indicates how many layers the frame has (max 4)"/>
+    </event>
+    <event name="receive_object">
+      <description summary="object receiving callback">
+        Callback which serves to supply the client with the file descriptors
+        containing the data for each object.
+        The amount of data in each FD can be inferred via lseek().
```
agx: At least weston does not assume that dmabufs are seekable:

https://github.com/wayland-project/weston/blob/master/libweston/linux-dmabuf.c#L228

atomnuker: Vulkan specifies that all mappable dmabuf file descriptors are seekable, but adding a field won't hurt. Done.

unstable/zwlr-screencontent-v1.xml
emersion: This one: https://wayland.freedesktop.org/docs/html/apa.html#protocol-spec-wl_shm-enum-format

Yeah, it is designed to exactly match `DRM_FORMAT_*`.

unstable/zwlr-screencontent-v1.xml
```diff
@@ -0,0 +1,232 @@
...
+      all the data has been made available for readout, as well as the time
+      at which presentation happened at.
+      The ownership of the frame is passed to the client, who's responsible for
+      destroying it once ready.
+      The data the API describes has been based off of what
+      VASurfaceAttribExternalBuffers contains.
+      All frames are read-only and may not be written into or altered.
+    </description>
+
+    <enum name="transform">
```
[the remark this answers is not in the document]
atomnuker: I can't seem to find a way to do that, wayland-scanner keeps complaining that the enum isn't found. How can I make wayland-scanner include the wayland.xml main protocol?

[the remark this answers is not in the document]
emersion: I checked and the string to use would be `wl_output.transform`. wayland-scanner still doesn't accept it, seems like a bug. I've sent an email to the wayland-devel ML.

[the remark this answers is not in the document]
SirCmpwn: https://lists.freedesktop.org/archives/wayland-devel/2018-April/037960.html

unstable/zwlr-screencontent-v1.xml
davidedmundson: This should probably be wayland pixel formats

(sircmpwn writing on this laptop)

emersion: Nah, this is different. This is a format _modifier_.

unstable/zwlr-screencontent-v1.xml
```diff
@@ -0,0 +1,220 @@
...
+      The ownership of the frame is passed to the client, who's responsible for
+      destroying it once ready.
+      The data the API describes has been based off of what
+      VASurfaceAttribExternalBuffers contains.
+      All frames are read-only and may not be written into or altered.
+    </description>
+
+    <enum name="frame_type">
+      <entry name="byte_buffer" value="0"
+             summary="file descriptors hold mappable linear memfd buffers"/>
```
ascent12: I don't really think this should specifically reference memfd as the way this is implemented, considering it's a Linux-only interface, and excludes other ways this could be implemented.
I think FreeBSD's SHM_ANON could be used to implement this, but I'm not certain.

unstable/zwlr-screencontent-v1.xml
```diff
@@ -0,0 +1,220 @@
...
+      <description summary="frame flags">
+        Special flags that must be respected by the client.
+        Transient frames indicate short lifetime frames (such as swapchain
+        images from external clients). Clients are advised to copy them and do
+        all processing outside of the "ready" event.
+      </description>
+      <entry name="transient" value="1"/>
+    </enum>
+
+    <event name="receive">
```
emersion: A more Wayland-y naming would be to remove "receive" and only put the name of the thing received.

emersion:
How would a command-line app wanting to screencast a specific wl_output request that output?

unstable/zwlr-screencontent-v1.xml
```diff
@@ -0,0 +1,220 @@
...
+    Overall usage:
+
+    1.) client registers with zwlr_screencontent_manager
+    2.) server sends client info about surfaces via "receive_surface_info"
+    3.) client subscribes to capture a surface via the "capture" request
+    4.) server sends client events via the "zwlr_screencontent_frame" interface
+    5.) client finishes and informs server via the "destroy" event
+  </description>
+
+  <interface name="zwlr_screencontent_frame" version="1">
```
emersion: Should add a `_v1` suffix

unstable/zwlr-screencontent-v1.xml
```diff
@@ -0,0 +1,220 @@
...
+
+    <request name="destroy" type="destructor">
+      <description summary="delete this object, used or not">
+        Unreferences the frame, allowing it to be reused. Must be called as soon
+        as its no longer used.
+      </description>
+    </request>
+  </interface>
+
+  <interface name="zwlr_screencontent_manager" version="1">
```
emersion: Same

unstable/wlr-screencontent-unstable-v1.xml
```diff
@@ -28,14 +28,14 @@
...
     2.) server sends client info about surfaces via "receive_surface_info"
     3.) client subscribes to capture a surface via the "capture" request
-    4.) server sends client events via the "zwlr_screencontent_frame" interface
-    5.) client finishes and informs server via the "destroy" event
+    4.) server sends client events via the "wlr_screencontent_frame" interface
+    5.) client finishes and informs server via the "frame_destroy" event
   </description>
 
-  <interface name="zwlr_screencontent_frame" version="1">
+  <interface name="wlr_screencontent_frame_v1" version="1">
```
emersion: You need to add a "z" prefix to mark the protocol as unstable (the protocol name doesn't have it because "unstable" is explicitly included).

unstable/wlr-screencontent-unstable-v1.xml
```diff
@@ -0,0 +1,220 @@
...
+    <event name="frame_abort">
+      <description summary="indicates the frame is no longer valid">
+        If the frame is no longer valid after the "receive" event has been
+        called, this callback will be used to inform the client to scrap the
+        frame. The client can register to receive another frame callback here
+        in case capturing becomes possible again.
+      </description>
+    </event>
+
+    <request name="frame_destroy" type="destructor">
```
emersion: Since thiese events and requests are in the `zwlr_screencontent_frame_v1` interface, no need to add a "frame_" prefix to those, otherwise we'll end up with symbols with "frame" appearing twice (like `zwlr_screencontent_frame_v1_frame_destroy(struct zwlr_screencontent_frame_v1 *frame)`).

unstable/wlr-export-dmabuf-unstable-v1.xml
```diff
@@ -0,0 +1,227 @@
...
+           summary="nanoseconds part of the timestamp"/>
+    </event>
+    <event name="frame_abort">
+      <description summary="indicates the frame is no longer valid">
+        If the frame is no longer valid after the "frame" event has been called,
+        this callback will be used to inform the client to scrap the frame.
+      </description>
+    </event>
+
+    <request name="frame_destroy" type="destructor">
```
emersion: Since these events and requests are in the `zwlr_screencontent_frame_v1` interface, no need to add a "frame_" prefix to those, otherwise we'll end up with symbols with "frame" appearing twice (like `zwlr_screencontent_frame_v1_frame_destroy(struct zwlr_screencontent_frame_v1 *frame)`).

unstable/wlr-export-dmabuf-unstable-v1.xml
```diff
@@ -0,0 +1,230 @@
...
+        been closed.
+      </description>
+      <arg name="id" type="uint"
+           summary="an identifier of the surface"/>
+    </event>
+
+    <request name="capture_client">
+      <description summary="subscribe to start capturing">
+        Request to start capturing from a surface with a given id.
+        If an ID becomes unavailable, a NULL will be returned.
```
emersion: It's not possible to return NULL because Wayland is an asynchronous protocol. Instead, you can return an abort frame.

unstable/wlr-export-dmabuf-unstable-v1.xml
```diff
@@ -0,0 +1,230 @@
...
+           summary="overlay custom client hardware cursor on top of the frame"/>
+    </request>
+    <request name="capture_output">
+      <description summary="subscribe to start capturing">
+        Request to start capturing from an entire wl_output.
+        If an output becomes unavailable, a NULL will be returned.
+      </description>
+      <arg name="frame" type="new_id"
+           interface="zwlr_export_dmabuf_frame_v1"/>
+      <arg name="out" type="object" interface="wl_output" allow-null="false"/>
```
emersion: `out` is a little vague, generally arguments that take an output are name `output`. `allow-null` is already set to `false` by default.

unstable/wlr-export-dmabuf-unstable-v1.xml
```diff
@@ -0,0 +1,230 @@
...
+    5.) client finishes and informs server via the "frame_destroy" event
+    6.) client optionally resubscribes via repeating steps 3.) through 5.)
+  </description>
+
+  <interface name="zwlr_export_dmabuf_frame_v1" version="1">
+    <description summary="a frame ready for readout">
+      This object represents a frame which is ready to have its resources
+      fetched and used.
+
+      The receive callback shall be called first, followed by either the
```
emersion: "either" makes it sound like "dma_object" might not be sent if "dma_layer" is sent once per layer

emersion:
We should add a `destroy` request for `zwlr_export_dmabuf_manager_v1`.

unstable/wlr-export-dmabuf-unstable-v1.xml
```diff
@@ -0,0 +1,230 @@
...
+      <arg name="mod_high" type="uint"
+           summary="drm format modifier, high"/>
+      <arg name="mod_low" type="uint"
+           summary="drm format modifier, low"/>
+      <arg name="num_objects" type="uint"
+           summary="indicates how many objects (FDs) the frame has (max 4)"/>
+      <arg name="num_layers" type="uint"
+           summary="indicates how many layers the frame has (max 4)"/>
+    </event>
+    <event name="dma_object">
```
emersion: Again, this gives symbols like `zwlr_export_dmabuf_frame_v1_send_dma_object`, with `dma` appearing twice

unstable/wlr-export-dmabuf-unstable-v1.xml
```diff
@@ -0,0 +1,230 @@
...
+    </description>
+
+    <enum name="frame_flags">
+      <description summary="frame flags">
+        Special flags that must be respected by the client.
+        Transient frames indicate short lifetime frames (such as swapchain
+        images from external clients). Clients are advised to copy them and do
+        all processing outside of the "ready" event.
+      </description>
+      <entry name="transient" value="1" since="1"/>
```
emersion: Do we need this at all?

unstable/wlr-export-dmabuf-unstable-v1.xml
```diff
@@ -167,6 +177,8 @@
         This may get called if for instance the surface is in the process of
         resizing.
       </description>
+      <arg name="reason" type="uint" enum="abort_reason"
+           summary="indicates reason"/>
```
SirCmpwn: This summary is not useful

emersion:
Linking KDE prior art here: https://cgit.kde.org/kwayland.git/commit/src/client/protocols/remote-access.xml

emersion:
Trying to summarize what's been said on IRC:
* @SirCmpwn said that we should remove the `source_output` `description` event, as this is already exposed by xdg-output and it's easy to add xdg-output support to a client. I agree.
* Do we want to expose surfaces or views? Views have subsurfaces and popups.
  - @SirCmpwn would like to send one source_surface per wl_surface, with a parent/child relationship including the position of the child relative to the parent. That would allow clients to filter each surface.
  - I would prefer to composite captured views on a separate DRM plane and send a DMA-BUF for this plane. If not enough overlay planes are available, it's also possible to use the primary plane and adjust the offset and the size of the DMA-BUF to crop it (but for transparent parts, it would include stuff below the view).
  - Yet another option would be to have a source_view event that yields multiple frames when captured (and specifies the location of frames relative to each other). An issue with @SirCmpwn's approach is that the relative position of surfaces is not fixed and can change over time - this solution would only give the position when frames are captured, removing race conditions.
* Should we stick to source IDs? @SirCmpwn has proposed to switch to server-side created objects. That would allow for a cleaner API, but there are some issues with lifecycle management that need to be understood before that (see http://ppaalanen.blogspot.co.uk/2014/07/wayland-protocol-design-object-lifespan.html)
* Should we use a dedicated protocol to list views and their surfaces? (#14)

Let me know if I forgot something or if there's a mistake in this summary.

atomnuker:
I disagree, honestly, not adding a single field will result in more than 100 lines more for clients, as they'd have to bind to the protocol, keep track of all outputs given by that protocol and then iterating and comparing all to figure out which one maps to which. I think this is absolutely ridiculous and justifies no need for any protocol interoperability.

SirCmpwn:
>I disagree, honestly, not adding a single field will result in more than 100 lines more for clients, as they'd have to bind to the protocol, keep track of all outputs given by that protocol and then iterating and comparing all to figure out which one maps to which. I think this is absolutely ridiculous and justifies no need for any protocol interoperability.

What you just described sounds like about 10 lines of code to me

unstable/wlr-export-dmabuf-unstable-v1.xml
```diff
@@ -0,0 +1,252 @@
...
+        The "transform" argument describes the orientation needed to be applied
+        to correctly orient the buffer. For example, a buffer rotated by 90
+        degrees will have a value of "3" here, corresponding to the need to
+        apply a 270 degree transpose to correctly present the buffer.
+      </description>
+      <arg name="width" type="uint"
+           summary="frame width, scaling factor included"/>
+      <arg name="height" type="uint"
+           summary="frame height, scaling factor included"/>
+      <arg name="scale" type="uint"
```
emersion: How can we handle fractional scaling?

I'm not a fan of making this a `fixed`.

For outputs, the client can read all the necessary transform information from wl_output and xdg-output. The integral scale and the transform are in wl_output, the logical size of the output (which gives a x- and a y-axis fractional scale) is in xdg-output. So maybe the client can do a better job using that instead of our frame properties.

We still need to support clients. Clients can only apply integral scaling and transformations to their buffers, not fractional scaling. But some other protocols (e.g. wp_viewporter) might add new transformations to client buffers, not sure how to handle those.

SirCmpwn:
If you're not willing to cooperate on this, I'm not willing to implement it. The protocol you have here is not going to work. Let me know when you're ready to have a discussion about my proposed changes. The changes we need to discuss are:

Planes and layers are done differently from dmabuf-import here. If there's no way to use dmabuf-import to make a client surface which would take advantage of your supposedly more correct design, then your more correct design is not useful. We need to either align this protocol with dmabuf-import or align dmabuf-import with this protocol.

We need to decide if the compositor is going to do a second compositing step(s) or if we want to let the capture client do the compositing. I prefer the latter approach because it means we never have to lie about anything, we never have to fill in a background which doesn't exist. Letting the capture client decide how to handle subsurfaces/popups which extent outside the bounds of the toplevel is also better. Exposing these surfaces separately also allows the capture client to choose not to render popups. Also, adding more compositing steps in the compositor could be very expensive. Even if we are able to render a captured surface in its own plane, we then have to have two more planes for everything below and everything above the compositor. If the client chooses to record two windows, then this just gets worse, and we quickly run up against hardware limitations. Capturing the entire output on top of that becomes even more complicated and now could require the client to composite everything anyway.

We should split the "new source" events into separate events for surfaces and outputs, and we should unify the capture request. Surfaces and outputs are different enough that we should not try to combine them into one, especially if we decide to give the capture client responsibility for compositing surfaces. If we also explicitly state that the compositor may respond with none, some, or all of the sources it has and does not guarantee sending them in a timely manner, we can also address the security question by giving the compositor time to prompt the user for which surfaces/outputs to share.

The frame resource is half designed with a short lifetime and half designed with a persistent lifetime. I like the persistent approach (where a single frame resource is reused for several frames) but we need to make changes to get it there. For one, we need to replace the destroy request with a finish request.

Multiple places where the descriptions which refer to designs which are no longer true need to be fixed. We need to elaborate on edge cases and specify the format and usage of each argument more clearly. We also need to align the style with conventions used by everyone else.

XML at https://github.com/atomnuker/wlr-protocols/pull/2

atomnuker:
On the subject of the frame structure:
I'm willing to work with you here, I don't mind some of the things you described, but I would like to stick with the layer abstraction for planes. Granted, it is for a fringe case (multiplanar images with multiple objects), but it allows simpler importing into VAAPI, which is the number one use case for this protocol. You can still feed in what you've been given into the dmabuf import protocol, though you're more limited in what you could give it. The layer abstraction is there to make code simpler by not having clients do heuristics to figure out how to organize planes. Just because the people behind the import protocol didn't think about this use case at the time (since they didn't have to) doesn't mean we should stick to what they made.

On the subject of API unification:
I think having 2 APIs for wl_outputs and client IDs is the lesser evil than needing to define a redundant event to give you wl_outputs, which you most likely already have. So I'd like to stick with this as well. Its just a single if () after all, and you have to use another protocol to define what description wl_outputs have.

On the subject of compositor compositing surfaces:
The idea is that if a client has no subsurfaces or anything but a single surface, the compositor will give that directly. Otherwise, the compositor will give the display dmabuf, offset and cropped to only show the single client.
The only case where this fails is if the client that's being captured is half-visible in the wl_output or the client itself is semi-transparent and you'd like for it to be fully opaque. In which case I'd rather leave handling unspecified. A compositor should be able to choose whether to do another step of compositing to a separate surface, or give the display dmabuf anyway with crop information so clients know which part to black out, or just give the base surface of the client without any subsurfaces.
Popups and subsurfaces could be exposed to be recorded separately. I don't have an opinion on this, I wouldn't mind defining relations between surfaces in the protocol. I can't see any use for this though, so most clients will just likely ignore popups from lists, which will require some extra info.

agx:
> I'm willing to work with you here, I don't mind some of the things you described, but I would like to stick
> with the layer abstraction for planes. Granted, it is for a fringe case (multiplanar images with multiple
> objects), but it allows simpler importing into VAAPI, which is the number one use case for this protocol. 

Can you give some code examples where layers as used in VAAPI are nicer?

SirCmpwn:
>Just because the people behind the import protocol didn't think about this use case at the time (since they didn't have to) doesn't mean we should stick to what they made.

No, but see:

>We should ask the linux-dmabuf authors if there's a good reason layers are not included in their protocol.

If you think the layer (or lack thereof) design in dmabuf import is wrong, you should sync up with them and make sure you know why they designed it that way and perhaps it can be improved.

Something that often happens in the wlroots project is the acknowledgement that our protocols don't live in a vacuum, but among their peers in an ecosystem of Wayland protocols. If we find issues, it's not okay to keep quiet, but to raise them and personally do the work necessary to improve the whole ecosystem.

>[lots of comments on the subject of surfaces]

Let's just forget about capturing individual surfaces in this version of the protocol and come back to it once we have a separate surface enumeration protocol (#14). To that end, I suppose I am more supportive of having two events: `new_output` and `capture_output` (explicitly named as such to allow us to distinguish them from other kinds of sources in the future, which may have different requirements). We don't have to have a source ID, which has always kind of sucked, we can just use one event to say that some `wl_output` is available for capture (which the client has to take note of in their own state somewhere) and then they can use the same `wl_output` to initiate a capture for it.

>The frame resource is designed to be used once, it's not half designed for one lifetime. The linux-dmabuf protocol is also using this approach (and in fact, all objects in the core wayland protocol and wayland-protocols are using it, none is re-usable). I think breaking this has no benefit and is inconsistent with the rest of the ecosystem.

So are you supposed to call `capture` at 60 Hz or whatever the framerate of the output is? Gross, I'd rather do it the way I suggested. If we really want to use short lifetime resources we could set up a wl_listener in the style of the surface frame event.

emersion:
Oh, sorry, I misunderstood. A wl_callback doesn't make sense here, and is inconsistent with linux-dmabuf. A wl_callback is used in the wl_surface interface because of races.

atomnuker:
A wl_callback is the way all frame callbacks operate, can't races happen here as well?

emersion:
>we can just use one event to say that some wl_output is available for capture

You want this event to let the compositor has a chance to ask the user which outputs he wants to share, but I'm not sure it's actually a good idea. With this solution, when a client binds to the manager the user will have to choose sources to share, and then the client will ask again the user what source to capture from. This double-prompt seems like bad user experience.

There are two other solutions:
1. Make the compositor responsible for choosing a source. The client would just be able to say "I want to capture" and the compositor will let the user choose the source that will be shared, or deny access (ala-HTML5).
2. Let the client request any source, and prompt the user for authorization when it starts capturing.

I'm not happy with any of these solutions.

atomnuker:
Solution 2 is fine, but since you're not happy with it you don't want any security?

emersion:
>Solution 2 is fine, but since you're not happy with it you don't want any security?

Nah, it just means I don't think this is a perfect solution, and I'll continue to think about it to try to find a better one. Among all three solutions, I prefer this one though.

unstable/wlr-export-dmabuf-unstable-v1.xml
```diff
@@ -0,0 +1,233 @@
...
+           summary="low 32 bits of the seconds part of the timestamp"/>
+      <arg name="tv_nsec" type="uint"
+           summary="nanoseconds part of the timestamp"/>
+    </event>
+
+    <enum name="abort_reason">
+      <description summary="abort reason">
+        Indicates reason for aborting the frame.
+      </description>
+      <entry name="invalid_source" value="0" since="1"
```
emersion: I need a reason to say "the DMA-BUF export failed, I'm not sure why, but you may try again and maybe it'll work".

Maybe we could have a boolean saying if the failure is temporary instead?

atomnuker:
Clients are meant to crop themselves at the end, and indeed often have to because hardware surfaces, particularly from hardware decoders, often have a non-constant amount of padding at a number of edges.

unstable/wlr-export-dmabuf-unstable-v1.xml
emersion: Let me reformulate my question: do we need the receiver to know the reason of the failure rather than to know if he needs to retry or not? Does this enum add any value to the protocol?

emersion:
>Clients are meant to crop themselves at the end, and indeed often have to because hardware surfaces, particularly from hardware decoders, often have a non-constant amount of padding at a number of edges.

So is it necessary to add these offset params, since the client will end up cropping stuff anyway, and all of this is zero-copy?

atomnuker:
Yes, because in case the compositor gives the client the display dmabuf but makes it track a single window the offset will need to be set to the offset of the window on the wl_output, and like we discussed on IRC, you can't use the dmabuf plane offset for tiled images, you have to signal the offset via external means.
dmabufs not from hardware decoders don't need to be cropped since they always start at 0 offset, but clients need to handle any amount of cropping in order to support hardware decoders hence why I mentioned adding it is not a problem.

emersion:
ACK, thanks for the explanation.

unstable/wlr-export-dmabuf-unstable-v1.xml
```diff
@@ -0,0 +1,244 @@
...
+      <arg name="transform" type="uint"
+           summary="transform needed to correct the orientation, has the same
+                    values as wl_output::transform"/>
+      <arg name="buffer_flags" type="uint"
+           summary="flags which indicate properties (invert, interlacing),
+                    has the same values as zwp_linux_buffer_params_v1:flags"/>
+      <arg name="flags" type="uint" enum="flags"
+           summary="indicates special frame features"/>
+      <arg name="format" type="uint"
+           summary="format of the layer (DRM_FORMAT_*)"/>
```
emersion: s/layer/frame/

unstable/wlr-export-dmabuf-unstable-v1.xml
```diff
@@ -0,0 +1,244 @@
...
+           summary="size in bytes for the current object"/>
+    </event>
+    <event name="plane">
+      <description summary="plane info receiving callback">
+        Callback which supplies the client with plane information for each
+        layer.
+      </description>
+      <arg name="index" type="uint"
+           summary="index of the current plane"/>
+      <arg name="layer_index" type="uint"
```
emersion: Should be removed

unstable/wlr-export-dmabuf-unstable-v1.xml
```diff
@@ -0,0 +1,244 @@
...
+        apply a 270 degree transpose to correctly present the buffer.
+      </description>
+      <arg name="width" type="uint"
+           summary="frame width, scaling factor included"/>
+      <arg name="height" type="uint"
+           summary="frame height, scaling factor included"/>
+      <arg name="offset_x" type="uint"
+           summary="crop offset for the x axis"/>
+      <arg name="offset_y" type="uint"
+           summary="crop offset for the y axis"/>
```
emersion: Security-wise, there's no way of enforcing these offsets?

emersion:
This version has wayland-scanner errors. (The build script should be updated to build all protocols, not just layer-shell)

unstable/wlr-export-dmabuf-unstable-v1.xml
atomnuker: Done

unstable/wlr-export-dmabuf-unstable-v1.xml
atomnuker: Done

unstable/wlr-export-dmabuf-unstable-v1.xml
atomnuker: Clients should sanity check when it makes sense (you've downloaded the frame and now you have to crop it) or do nothing (you're rendering the frame to some framebuffer and all that can happen is its not going to be visible). Its the same as any other part of the protocol (check the fds, check the format, check the modifier, etc.).

emersion: Nah, I mean: if the client asks to capture another client, and the compositor decides to crop the DMA-BUF of the whole screen with these offsets, the client can ignore those and read pixels it shouldn't have access to.

emersion:
Let's try to summarize what remains to be done here:

* Parts of the protocol allowing to capture other clients should be removed (per https://github.com/swaywm/wlr-protocols/pull/11#issuecomment-392038387)
* The wording needs improvement (see @SirCmpwn's PR)

Minor issues:

* How to handle fractional scaling? (see https://github.com/swaywm/wlr-protocols/pull/11#discussion_r190529514)
* I'm not a fan of the `abort_reason` enum (see https://github.com/swaywm/wlr-protocols/pull/11#discussion_r191013245)

Let me know if you don't agree with something or if I forgot something.

atomnuker:
Yeah, that sounds fine, I'll do that instead.

emersion:
>We need some way to explain to the clients why capturing failed.

But why? Only saying to them "this failure is temporary" or "this failure is permanent" should be enough.

>As for wording, incremental PRs accepted rather than a large dump which changes both formatting and text

+1

>Fractional scaling should probably be covered by cross-referencing this with xdg-output

+1, but then do we need scale and transform as arguments here?

SirCmpwn:
>Only saying to them "this failure is temporary" or "this failure is permanent" should be enough

+1

SirCmpwn:
>+1, but then do we need scale and transform as arguments here?

I would say yes

atomnuker:
We do need scale as argument here if we're capturing clients. Fractional scaling can be deduced from wl_output and xdg_output into.
Additionally, clients may have a different transform than outputs, say for example in an image viewer presenting transposed images.

emersion:
>We do need scale as argument here if we're capturing clients.

I don't see why output sources should get this information from the wl_output and client sources should get it from here. We can just add scale/transform to the other protocol.

atomnuker:
Fine, I'll remove them both, I really have no opinion or use for them, I just added them for completeness sake.

atomnuker:
Or maybe I shouldn't, because they won't be frame perfect, as the order in which events from different protocols are sent is not specified.

emersion:
>Fine, but what about the transform?

You get it from wl_output too.

SirCmpwn:
Frame perfect is imo more important than DRY. In this case it makes sense to me to attach information about how to interpret the buffer (such as scale factor and transform) to the events here.

However, we should just stick to xdg output for dealing with fractional scaling. Once you have fractional scaling it's impossible to render perfectly anyway, so fuck it.

emersion:
How do you deal with the output position then? Do you just copy them over here?

SirCmpwn:
Is anyone really going to be compositing several outputs together? Do we really care if the position is frame perfect so long as the pixels are correct?

atomnuker:
I agree, frame perfect reproduction is really necessary for this protocol, otherwise we'd look like complete amateurs and have something no better than what's already out there.

emersion:
I think adding half of the output metadata doesn't make sense. These aren't properties about the buffer or the frame, these are properties about the output itself.

SirCmpwn:
Hmm, can libav et al even handle changing the output resolution in response to a changing scale factor or transform?

emersion:
Note that the compositor can just send "resizing" if the output scale changes between the capture request (ie. when the frame/plane events are sent) and the ready event.

atomnuker:
Yep, it can do it for the transform and the output resolution, it can't for the scale since its never been needed for anything. In fact the only container which can signal the scale is mov, with a custom apple-specific scaling factor extension, which lavf still can't write or read.
So I wouldn't mind getting rid of the scale, the transform is what's needed to be frame perfect.

SirCmpwn:
>Note that the compositor can just send "resizing" if the output scale changes between the capture request (ie. when the frame/plane events are sent) and the ready event.

This is a good point, we can just make sure everything is synced up in the compositor and simplify this protocol

atomnuker:
That could work.

atomnuker:
Though I'd still like for the transform to be included in the frame event, because otherwise the client will need to queue it up for the next or current frame and output the correct side data at the correct time.

atomnuker:
Removed scale and transform

unstable/wlr-export-dmabuf-unstable-v1.xml
```diff
@@ -0,0 +1,201 @@
...
+           summary="nanoseconds part of the timestamp"/>
+    </event>
+
+    <enum name="cancel_reason">
+      <description summary="cancel reason">
+        Indicates reason for cancelling the frame.
+      </description>
+      <entry name="temporary" value="0" since="1"
+             summary="temporary error, source will produce more frames"/>
+      <entry name="pernament" value="1" since="1"
```
emersion: Typo: permanent

unstable/wlr-export-dmabuf-unstable-v1.xml
```diff
@@ -0,0 +1,201 @@
...
+        as its no longer used.
+        Can be called at any time by the client after the "frame" event, after
+        which the compositor will not call any other events unless the client
+        resubscribes to capture more. The client will still have to close any
+        FDs it has been given.
+      </description>
+    </request>
+  </interface>
+
+  <interface name="zwlr_export_dmabuf_manager_v1" version="1">
```
emersion: No need to specify "version" if it's 1.

unstable/wlr-export-dmabuf-unstable-v1.xml
```diff
@@ -0,0 +1,201 @@
...
+      responsible for destroying it via the "destroy" event once finished and
+      by calling close() on the file descriptors received.
+      All frames are read-only and may not be written into or altered.
+    </description>
+
+    <enum name="flags">
+      <description summary="frame flags">
+        Special flags that should be respected by the client.
+      </description>
+      <entry name="transient" value="0x1" since="1"
```
emersion: No need to specify "since" if it's 1.

unstable/wlr-export-dmabuf-unstable-v1.xml
```diff
@@ -0,0 +1,201 @@
...
+  <description summary="a protocol for low overhead screen content capturing">
+    An interface to capture surfaces in an efficient way.
+    Overall usage:
+
+    1.) client registers with zwlr_screencontent_manager_v1
+    2.) server sends client info about surfaces via "receive_surface_info"
+    3.) client subscribes to capture a surface via the "capture" requests
+    4.) server sends client events via the "zwlr_screencontent_frame" interface
+    5.) client finishes and informs server via the "frame_destroy" event
+    6.) client optionally resubscribes via repeating steps 3.) through 5.)
```
SirCmpwn: This comment is outdated

unstable/wlr-export-dmabuf-unstable-v1.xml
```diff
@@ -0,0 +1,201 @@
...
+    <event name="frame">
+      <description summary="main callback">
+        Main callback supplying the client with information about the frame,
+        as well as an object to serve as context for destruction. Always called
+        first before any other events.
+
+        The "transform" argument describes the orientation needed to be applied
+        to correctly orient the buffer. For example, a buffer rotated by 90
+        degrees will have a value of "3" here, corresponding to the need to
+        apply a 270 degree transpose to correctly present the buffer.
```
SirCmpwn: This comment also outdated

unstable/wlr-export-dmabuf-unstable-v1.xml
```diff
@@ -0,0 +1,201 @@
...
+      </description>
+      <arg name="width" type="uint"
+           summary="frame width, scaling factor included"/>
+      <arg name="height" type="uint"
+           summary="frame height, scaling factor included"/>
```
SirCmpwn: I would specify that this is in pixels instead of referencing a scale factor

unstable/wlr-export-dmabuf-unstable-v1.xml
```diff
@@ -0,0 +1,201 @@
...
+      <arg name="frame" type="new_id"
+           interface="zwlr_export_dmabuf_frame_v1"/>
+      <arg name="overlay_cursor" type="int"
+           summary="include custom client hardware cursor on top of the frame"/>
+      <arg name="output" type="object" interface="wl_output"/>
+    </request>
+    <request name="destroy" type="destructor">
+      <description summary="destroy the manager">
+        All objects created by the manager will still remain valid, until their
+        appropriate destroy request has been called.
```
SirCmpwn: Typically it's preferred to make destroying the parent interface a protocol error if any of its child resources are still alive

unstable/wlr-export-dmabuf-unstable-v1.xml
```diff
@@ -0,0 +1,201 @@
...
+    </description>
+
+    <request name="capture_output">
+      <description summary="subscribe to start capturing">
+        Request to start capturing from an entire wl_output.
+      </description>
+      <arg name="frame" type="new_id"
+           interface="zwlr_export_dmabuf_frame_v1"/>
+      <arg name="overlay_cursor" type="int"
+           summary="include custom client hardware cursor on top of the frame"/>
```
SirCmpwn: I think we should remove this

unstable/wlr-export-dmabuf-unstable-v1.xml
emersion: Hmm, not really. For instance the linux-dmabuf protocol also allows this. Same for wl_compositor, wl_subcompositor…

SirCmpwn:
The desire to avoid reusing the frame resource combined with the semantics of the cancel event don't make any sense to me. Why would we need to distinguish between temporarily cancelled and permenantely cancelled in this case?

unstable/wlr-export-dmabuf-unstable-v1.xml
SirCmpwn: The example I was looking at is xdg-shell. But if there's no consensus then forget it, this is fine.

unstable/wlr-export-dmabuf-unstable-v1.xml
emersion: Yeah, I'm still not sure how to implement it anyway. @ascent12, thoughts?

emersion:
>Why would we need to distinguish between temporarily cancelled and permenantely cancelled in this case?

If the failure is temporary, it means the export failed, but calling again `capture_output` with the same output may work in the future (e.g. no more FDs or some other stupid reason). If the failure is permanent, it means it won't ever work to capture this output (e.g. the backend doesn't support it).

A frame will either receive a `ready` or a `cancel` event, and won't ever be used twice.

SirCmpwn:
I think that should be described like you just said in the docs. I'm not fond of the numbered list of steps, better to explain it in terms like those.

unstable/wlr-export-dmabuf-unstable-v1.xml
atomnuker: Absolutely not, this is a required feature for many people who'd like to stream. Like I said, you can use the UNIVERSAL PLANES flag to get the final composited dmabuf which will have the cursor.

SirCmpwn: Sure, we can get the buffer without the cursor. But getting the buffer without the cursor requires an extra compositing step if we have to fall back to software cursors even before dmabuf-export is considered.

SirCmpwn: Actually, getting it with the cursor might also be nebulous, not all drivers support universal planes either.

atomnuker: Fine if the description is changed to say that this might be inconsistent? It can't be made consistent (because clients can have their own soft cursor) but this option does need to stay as for the majority of cases it'll work fine.
Software cursors only happen in obscure edge cases anyway, and they're usually the cases where you want to have them overlaid.

SirCmpwn: I have a laptop which doesn't support hardware cursors at all, it's not that obscure.

For v1 of this protocol I'd rather just skip this option and let the compositor do whatever it wants.

atomnuker: I disagree, this reasoning is exactly why it took so long to have a stable and usable xdg_shell. And just because your laptop doesn't support it is no reason why the protocol has to be crippled. The bloody windows capture API has support for this as well, and it works in much the same way, if there's a software cursor it stays, otherwise you can choose between having a hardware cursor overlaid or not.

atomnuker: Okay, so lets leave it up to the compositor then?

atomnuker: Hmm, maybe not, I still think that if the compositor can it should respect the choice. If it can't, it'll just output whatever it can. Document this as such?

emersion: I'm personally okay with that. It's not perfect, but the compositor will have a best effort strategy.

(We don't _need_ universal planes for this, we can just force software cursors if the client wants cursors, and (try to) use hardware cursors if not. It gets messy when multiple clients start capturing the same output though…)

atomnuker: How so?

emersion: If a client requests the output with the cursor and another one requests the output without it, we can't do both. That could happen if you're using a RDP client and screencasting at the same time for instance.

unstable/wlr-export-dmabuf-unstable-v1.xml
```diff
@@ -0,0 +1,213 @@
...
+           summary="drm format modifier, high"/>
+      <arg name="mod_low" type="uint"
+           summary="drm format modifier, low"/>
+      <arg name="num_objects" type="uint"
+           summary="indicates how many objects (FDs) the frame has (max 4)"/>
+      <arg name="num_planes" type="uint"
+           summary="indicates how many planes the frame has (max 4)"/>
+    </event>
+
+    <event name="object">
```
emersion: The existing linux-dmabuf protocol doesn't have an object/plane distinction. This doesn't seem absolutely necessary since the FDs will be duplicated anyway. Is there a reason why we should keep them separate?

PR #17, opened 2018-05-23
Add screencopy protocol

### LINKED ISSUE #16: A better screenshooter protocol
We're building a protocol for DMA-BUF exporting (#11). In case it's not possible to use it (because the backend doesn't support it, eg. Wayland/X11/headless), we need a fallback which involves copying to a client buffer. Currently this is done with an old version of the Orbital screenshooter protocol. However, this doesn't let the client request to capture other clients or regions. Also the buffer setup is a little hacky because it requires the client to guess the buffer settings from the output mode. The protocol doesn't let us add flags (like "inverted_y") that would speed up the copy operation.

Since the outdated protocol doesn't seem to be used by anyone except wlroots, I'd like to improve it and improve the integration with the DMA-BUF export protocol to make automatic fallback easier.

Prior art:
* Orbital: https://github.com/giucam/orbital/blob/master/protocol/screenshooter.xml
* ~~KDE (it's closer to our DMA-BUF export interface, but with GBM FDs): https://github.com/KDE/kwayland/blob/master/src/client/protocols/remote-access.xml~~

emersion:
- [x] Capture outputs
- [x] Copy a region instead of the whole output
- [x] ~~Capture views (waiting to know #11's approach)~~ not yet
- [x] ~~Capture other stuff (e.g. sway workspaces/containers)~~

Fixes #16

### DIFF
--- a/unstable/wlr-screencopy-unstable-v1.xml
+++ b/unstable/wlr-screencopy-unstable-v1.xml
@@ -76,10 +76,6 @@
       destroy the frame.
     </description>
 
-    <enum name="flags" bitfield="true">
-      <entry name="y_invert" value="1" summary="contents are y-inverted"/>
-    </enum>
-
     <event name="buffer">
       <description summary="buffer information">
         Provides information about the frame's buffer. This event is sent once
@@ -91,7 +87,6 @@
       </description>
       <arg name="width" type="uint" summary="buffer width"/>
       <arg name="height" type="uint" summary="buffer height"/>
-      <arg name="flags" type="uint" enum="flags" summary="buffer flags"/>
       <arg name="format" type="uint" summary="preferred DRM_FORMAT"/>
       <arg name="stride" type="uint" summary="preferred stride"/>
     </event>
@@ -115,6 +110,18 @@
       <entry name="invalid_dimensions" value="2"
         summary="invalid width or height"/>
     </enum>
+
+    <enum name="flags" bitfield="true">
+      <entry name="y_invert" value="1" summary="contents are y-inverted"/>
+    </enum>
+
+    <event name="flags">
+      <description summary="frame flags">
+        Provides flags about the frame. This event is sent once before the
+        "ready" event.
+      </description>
+      <arg name="flags" type="uint" enum="flags" summary="frame flags"/>
+    </event>
 
     <event name="ready">
       <description summary="indicates frame is available for reading">
--- a/unstable/wlr-screencopy-unstable-v1.xml
+++ b/unstable/wlr-screencopy-unstable-v1.xml
@@ -26,6 +26,15 @@
   <description summary="screen content capturing on client buffers">
     This protocol allows clients to ask the compositor to copy part of the
     screen content to a client buffer.
+
+    Warning! The protocol described in this file is experimental and
+    backward incompatible changes may be made. Backward compatible changes
+    may be added together with the corresponding interface version bump.
+    Backward incompatible changes are done by bumping the version number in
+    the protocol and interface names and resetting the interface version.
+    Once the protocol is to be declared stable, the 'z' prefix and the
+    version number in the protocol and interface names are removed and the
+    interface version number is reset.
   </description>
 
   <interface name="zwlr_screencopy_manager_v1" version="1">
@@ -36,9 +45,11 @@
 
     <request name="capture_output">
       <description summary="start capturing">
-        Request to start capturing from the given output.
+        Capture the next frame of a an entire output.
       </description>
       <arg name="frame" type="new_id" interface="zwlr_screencopy_frame_v1"/>
+      <arg name="overlay_cursor" type="int"
+        summary="include custom client hardware cursor on top of the frame"/>
       <arg name="output" type="object" interface="wl_output"/>
     </request>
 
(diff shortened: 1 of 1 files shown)


unstable/wlr-screencopy-unstable-v1.xml
```diff
@@ -0,0 +1,141 @@
...
+    <request name="copy">
+      <description summary="copy the frame">
+        Copy the frame to the supplied buffer. The buffer must have a the
+        correct size, see zwlr_screencopy_frame_v1.buffer. The buffer needs to
+        have a supported format.
+
+        If the frame is successfully copied, a ready event is sent. Otherwise,
+        an abort event is sent.
+      </description>
+      <arg name="buffer" type="object" interface="wl_buffer"/>
```
emersion: Not sure we want to use a `wl_buffer` for this - maybe a FD would be more appropriate.

SirCmpwn: wl_buffer seems fine to me

SirCmpwn:
>Copy a region instead of the whole output

I would like to see this get in, and I would like to see a timestamp added telling you the time the frame you got captured was presented.

unstable/wlr-screencopy-unstable-v1.xml
```diff
@@ -0,0 +1,180 @@
...
+    <request name="capture_output_region">
+      <description summary="capture an output's region">
+        Capture the next frame of an output's region.
+
+        The region is given in output logical coordinates. See
+        xdg_output.logical_size.
+      </description>
+      <arg name="frame" type="new_id" interface="zwlr_screencopy_frame_v1"/>
+      <arg name="overlay_cursor" type="int"
+        summary="include custom client hardware cursor on top of the frame"/>
```
SirCmpwn: I don't think we need to get so specific here. `summary="composite cursor onto the frame"`

unstable/wlr-screencopy-unstable-v1.xml
```diff
@@ -0,0 +1,186 @@
...
+      <arg name="output" type="object" interface="wl_output"/>
+    </request>
+
+    <request name="capture_output_region">
+      <description summary="capture an output's region">
+        Capture the next frame of an output's region.
+
+        The region is given in output logical coordinates, see
+        xdg_output.logical_size. Trying to capture a region spanning outside the
+        output extents is a protocol error.
```
Ongy: This is racy against anything that changes logical size, resulting in a protocol error sounds sub-optimal to me.
The frame event has it's own size either way, is there a reason you don't want to clip it?

emersion: Hmm, I just wanted to avoid frames with no content. But you're right, it's a bad idea.

emersion:
>I can't imagine a situation where I'm going to be willing to do format conversions in the compositor for the sake of a picky client.

That's true, but in the case of OpenGL, the real format used in the GPU buffer is opaque and driver-specific, so we need to do format conversions anyway. It does make more sense to force the client to use the format we pick though.

emersion:
Ping?

SirCmpwn:
:+1:

unstable/wlr-screencopy-unstable-v1.xml
```diff
@@ -0,0 +1,179 @@
...
+    Once the protocol is to be declared stable, the 'z' prefix and the
+    version number in the protocol and interface names are removed and the
+    interface version number is reset.
+  </description>
+
+  <interface name="zwlr_screencopy_manager_v1" version="1">
+    <description summary="manager to inform clients and begin capturing">
+      This object is a manager which offers requests to start capturing from a
+      source.
+    </description>
```
agx: Would it make sense to differentiate this from the screenshoter and export-dmabuf protocols?

atomnuker:
I don't get why you want clients to supply their own buffer. Just leave it to clients to get an allocated buffer and refcount it themselves if they have to.

emersion:
>Just leave it to clients to get an allocated buffer 

What do you mean?

>Why are ready and flags two events? I couldn't figure that out from the commit message.

I wanted not to pollute the ready event with flags.

>Untransformed is IMHO o.k. since the client can get the transform from the output but it should be probably be noted in the protocol

Yeah, that's how clients are supposed to do. Same for scale.

agx:
On Fri, Jun 29, 2018 at 01:46:56AM -0700, emersion wrote:
[..snip..] 
>      Untransformed is IMHO o.k. since the client can get the transform from
>      the output but it should be probably be noted in the protocol
> 
>    Yeah, that's how clients are supposed to do. Same for scale.

Thought so. I'd just add a note to the protocol description about it.

atomnuker:
Can't clients directly get a wl_buffer without the request?

emersion:
>Can't clients directly get a wl_buffer without the request?

No, they can't read a compositor-supplied wl_buffer. wl_buffer is designed to be client-side allocated.

unstable/wlr-screencopy-unstable-v1.xml
SirCmpwn: If we differentiate this it would probably be at the export-dmabuf level, since the latter is more specific. As for screenshooter I think we can consider screenshooter obsoleted by this protocol.

PR #23, opened 2018-07-13
[WIP] Add output manager protocol

### LINKED ISSUE #15: Add a protocol to change output properties
It should be possible to change output resolution, scaling and rotation via a wayland protocol so that clients like gnome-control-center can do so without private protocol extensions. We could start simple with output transforms and scaling and refine the protocol in later revisions.

VincentVanlaer:
Fixes #15 
I still need to do a some descriptions and a final spell/grammar check, but the protocol itself can be reviewed.
The protocol references the wl_output.transform enum, so it will only build on the alpha that was just released.

Prior art by KDE:
https://cgit.kde.org/kwayland.git/tree/src/client/protocols/outputdevice.xml
https://cgit.kde.org/kwayland.git/tree/src/client/protocols/output-management.xml

Difference with the KDE protocols
- One protocol instead of two. No preference either way, one protocol was just easier to write.
- Removed subpixel. Doesn't seem very relevant for configuring outputs.
- Removed make & model. Replaced by name & description.
- Removed phys size. Not sure whether it should be added or not, would be useful to have a constant dpi across all output devices.
- Removed EDID. Including EDID feels like an overkill.
- Removed uuid. Name + description should be sufficient. Besides, I don't think its a good idea to force compositors to remember stuff across restarts for a protocol like this.
- Removed scale. Replaced by size.
- Removed requirement that outputs should not overlap or cause gaps. This is i.m.o. compositor dependent. 
- Added name & description.
- Added size. Similar to xdg-output's logical_size event. Allows for fractional scaling.
- Added disconnected. Outputs can be removed.
- Added destructors.
- Added the requirement that the server should try to rollback on failures.
- Added the requirement that only property changes requested explicitly by the client are applied. this improve consistency across compositors.
- Split mode into supported_mode and active_mode. Allows a client to set a custom mode.

Questions I still have:
- On a failed event, should we inform the client what caused the failure? Maybe an enum describing which request failed and the output the configuration change was applied to.
- What's the consensus on the copyright? Some descriptions are from the core protocol, the KDE protocol and xdg-output. I assume I need to add the copyright holders from those protocols as well?

### DIFF
--- a/unstable/wlr-output-management-unstable-v1.xml
+++ b/unstable/wlr-output-management-unstable-v1.xml
@@ -67,7 +67,11 @@
 
         When the client binds to the output_device object, the server sends this
         event once for every supported mode.
-        
+
+        If the concept of a specific refresh rate does not apply to the mode,
+        the refresh argument shpuld be zero. An example of this is a monitor
+        with a variable refresh rate.
+
         Not all output devices need to have supported modes. For example, an
         RDP-style output might have no advertised modes.
 
--- a/unstable/wlr-output-management-unstable-v1.xml
+++ b/unstable/wlr-output-management-unstable-v1.xml
@@ -69,7 +69,7 @@
         event once for every supported mode.
 
         If the concept of a specific refresh rate does not apply to the mode,
-        the refresh argument shpuld be zero. An example of this is a monitor
+        the refresh argument should be zero. An example of this is a monitor
         with a variable refresh rate.
 
         Not all output devices need to have supported modes. For example, an
--- a/unstable/wlr-output-management-unstable-v1.xml
+++ b/unstable/wlr-output-management-unstable-v1.xml
@@ -151,16 +151,6 @@
         summary="y position within the global compositor space"/>
     </event>
     
-    <enum name="enablement">
-      <description summary="describes enabled state">
-        Describes whether a device is enabled, i.e. device is used to
-        display content by the compositor. This wraps a boolean around
-        an int to avoid a boolean trap.
-      </description>
-      <entry name="disabled" value="0"/>
-      <entry name="enabled" value="1"/>
-    </enum>
-
     <event name="enabled">
       <description summary="output is enabled or disabled">
         This event describes whether the output is actively displaying a part
@@ -172,7 +162,8 @@
         This event is sent after the client binds to the output_device
         and whenever the output device is enabled or disabled.
       </description>
-      <arg name="enabled" type="int" summary="output enabled state"/>
+      <arg name="enabled" type="int"
+        summary="output enabled state (nonzero if enabled)"/>
     </event>
 
     <event name="name">
@@ -303,7 +294,8 @@
       <description summary="enable or disable the output">
         Enable or disable an output device.
       </description>
-      <arg name="enabled" type="int" summary="output enabled state"/>
+      <arg name="enabled" type="int"
+        summary="output enabled state (nonzero if enabled)"/>
       <arg name="output_device" type="object" interface="zwlr_output_device_v1"
         summary="the output device to be configured"/>
     </request>
--- a/unstable/wlr-output-management-unstable-v1.xml
+++ b/unstable/wlr-output-management-unstable-v1.xml
@@ -244,7 +244,7 @@
       <description summary="create a new output configuration object">
         Create a new output configuration object.
       </description>
-      <arg name="id" type="new_id" interface="zwlr_output_device_v1"/>
+      <arg name="id" type="new_id" interface="zwlr_output_configuration_v1"/>
     </request>
 
     <request name="destroy" type="destructor">
--- a/unstable/wlr-output-management-unstable-v1.xml
+++ b/unstable/wlr-output-management-unstable-v1.xml
@@ -345,6 +345,8 @@
         server should revert any changes made by the apply request that
         triggered this event.
       </description>
+      <arg name="reason" type="string"
+        summary="describes the reason for the failure"/>
     </event>
 
     <request name="destroy" type="destructor">
--- a/unstable/wlr-output-management-unstable-v1.xml
+++ b/unstable/wlr-output-management-unstable-v1.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<protocol name="wlr_output_manager_v1">
+<protocol name="wlr_output_management_unstable_v1">
   <copyright>
     Copyright © 2018 Vincent Vanlaer
 
--- a/unstable/wlr-output-management-unstable-v1.xml
+++ b/unstable/wlr-output-management-unstable-v1.xml
@@ -308,6 +308,18 @@
         summary="the output device to be configured"/>
     </request>
 
+    <request name="set_size">
+      <description summary="set the size of an output">
+        Set the size of an output device in the global compositor space.
+      </description>
+      <arg name="width" type="int"
+        summary="width in global compositor space"/>
+      <arg name="height" type="int"
+        summary="height in global compositor space"/>
+      <arg name="output_device" type="object" interface="zwlr_output_device_v1"
+        summary="the output device to be configured"/>
+    </request>
+
     <request name="apply">
       <description summary="apply pending changes">
         Apply all pending changes requested by the client through this
--- a/unstable/wlr-output-manager-v1.xml
+++ b/unstable/wlr-output-manager-v1.xml
@@ -95,7 +95,7 @@
       <arg name="refresh" type="int" summary="vertical refresh rate in mHz"/>
     </event>
     
-    <event name="transformation">
+    <event name="transform">
       <description summary="transformations applied to framebuffer">
         This event describes the transformations that are applied to the
         framebuffer before it is displayed.
@@ -103,7 +103,7 @@
         This event is sent after the client binds to the output_device and
         whenever the mode of the output changes.
       </description>
-      <arg name="transformation" type="int" enum="wl_output.transform"/>
+      <arg name="transform" type="int" enum="wl_output.transform"/>
     </event>
     
     <event name="size">
@@ -116,7 +116,7 @@
         will not be the same if the compositor does not scale the surface
         buffers.
 
-        Combined with the current_mode and transformation event, this event
+        Combined with the current_mode and transform event, this event
         effectively describes the scale applied to the output's buffer
         before it is displayed.
 
@@ -266,11 +266,11 @@
         summary="the output device to be configured"/>
     </request>
 
-    <request name="set_transformation">
+    <request name="set_transform">
       <description summary="set the buffer transformation">
         Set the buffer transformation of an output device.
       </description>
-      <arg name="transformation" type="int" enum="wl_output.transform" 
+      <arg name="transform" type="int" enum="wl_output.transform"
         summary="transformation to apply"/>
       <arg name="output_device" type="object" interface="zwlr_output_device_v1"
         summary="the output device to be configured"/>
--- a/unstable/wlr-output-manager-v1.xml
+++ b/unstable/wlr-output-manager-v1.xml
@@ -219,19 +219,6 @@
       </description>
     </event>
 
-    <event name="disconnected">
-      <description summary="">
-        The disconnected event indicates that the output device is no longer
-        connected to the compositor. The server will no longer send events to
-        this object.
-
-        If the client tries to configure any of the output's properties, it will
-        fail.
-
-        The client should destroy the object after it receives this event.
-      </description>
-    </event>
-
     <request name="destroy" type="destructor">
       <description summary="destroy the output device object">
         Using this request a client can tell the server that it is not going to
--- a/unstable/wlr-output-manager-v1.xml
+++ b/unstable/wlr-output-manager-v1.xml
@@ -179,13 +179,16 @@
 
         The naming convention is compositor defined, but limited to
         alphanumeric characters and dashes (-). Each name is unique among all
-        wl_output globals, but if a wl_output global is destroyed the same name
-        may be reused later. The names will also remain consistent across
-        sessions with the same hardware and software configuration.
+        output_device globals, but if a output_device global is destroyed the
+        same name may be reused later. The names will also remain consistent
+        across sessions with the same hardware and software configuration.
 
         Examples of names include 'HDMI-A-1', 'WL-1', 'X11-1', etc. However, do
         not assume that the name is a reflection of an underlying DRM
         connector, X11 connection, etc.
+
+        If the compositor implements xdg_output and this output is enabled,
+        the zxdg_output_v1.name must report the same name.
 
         This event is only sent once per zwlr_output_device_v1, and the
         name does not change over the lifetime of the zwlr_output_device_v1.
@@ -196,12 +199,15 @@
     <event name="description">
       <description summary="human-readable description of this output">
         Many compositors can produce human-readable descriptions of their
-        outputs.  The client may wish to know this description as well, to
+        outputs. The client may wish to know this description as well, to
         communicate the user for various purposes.
 
         The description is a UTF-8 string with no convention defined for its
         contents. Examples might include 'Foocorp 11" Display' or 'Virtual X11
         output via :1'.
+
+        If the compositor implements xdg_output and this output is enabled,
+        the zxdg_output_v1.name must report the description. 
 
         This event is only sent once per zwlr_output_device_v1, and the
         name does not change over the lifetime of the zwlr_output_device_v1.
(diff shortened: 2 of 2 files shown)


unstable/wlr-output-manager-v1.xml
```diff
@@ -0,0 +1,350 @@
...
+      <description summary="sent all information about output">
+        This event is sent after all other properties have been
+        sent after binding to the output object and after any
+        other property changes done after that. This allows
+        changes to the output properties to be seen as
+        atomic, even if they happen via multiple events.
+      </description>
+    </event>
+
+    <event name="disconnected">
```
emersion: Since this is a global, we can just remove the global instead.

unstable/wlr-output-manager-v1.xml
```diff
@@ -0,0 +1,350 @@
...
+        This event describes whether the output is actively displaying a part
+        of the global compositor space.
+
+        If an output_device is enabled, the output must be represented by
+        a wl_output.
+        
+        This event is sent after the client binds to the output_device
+        and whenever the output device is enabled or disabled.
+      </description>
+      <arg name="enabled" type="int" summary="output enabled state"/>
```
emersion: Instead of having an enum, we can just have an `int` which is non-zero when enabled

Ongy: Or something like:
`
<event name="wl_output>
<arg name="output" type="object" interface="wl_output" nullable/>
<event>
`

non-null when it's active, and null when sent to inactivity.
A direct link to wl_output (avoiding string comparisons) is missing either way IMO

Ongy:
> Removed uuid. Name + description should be sufficient. Besides, I don't think its a good idea to force compositors to remember stuff across restarts for a protocol like this.

I think we should try to get this used by KDE as well, might be nicer to have it as optional output attribute for those that want to support it for that purpose (and in general)

unstable/wlr-output-manager-v1.xml
```diff
@@ -0,0 +1,350 @@
...
+
+    <event name="name">
+      <description summary="name of this output">
+        Many compositors will assign names to their outputs, show them to the
+        user, allow them to be configured by name, etc. The client may wish to
+        know this name as well to offer the user similar behaviors.
+
+        The naming convention is compositor defined, but limited to
+        alphanumeric characters and dashes (-). Each name is unique among all
+        wl_output globals, but if a wl_output global is destroyed the same name
```
Ongy: `wl_output` globals? :)

I think you copy+pasted a bit eagerly.

Should probably reference the `xdg-output` name event (and force equality?).

Ongy:
> On a failed event, should we inform the client what caused the failure? Maybe an enum describing which request failed and the output the configuration change was applied to.

I'm leaning towards string. Enum could be annoying, and if we want to include e.g. "leaves gaps" which is allowed by the protocol, but might be imposed by the compositor, it gets ugly.

> What's the consensus on the copyright? Some descriptions are from the core protocol, the KDE protocol and xdg-output. I assume I need to add the copyright holders from those protocols as well?

As always: IANAL

Copyright is a mess. You didn't base this on the core, you just re-used necessary parts, afaik no need to consider that.
On the KDE side, did you base it on theirs? Otherwise adding them, without them ever seeing this is iffy as well IMO. In the long run we should probably get together with them either way, so this shouldn't be a huge problem.

emersion:
>I think we should try to get this used by KDE as well, might be nicer to have it as optional output attribute for those that want to support it for that purpose (and in general)

They can just use "name" for this.

Ongy:
Or rather description, yes. They didn't have this field before, and I'd bind the name to xdg-output name, as mentioned before

ammen99:
@VincentVanlaer Why did you use "transformation" and not "transform" throughout the protocol? The latter is used in `wl_output` so to me this seems an inconsistency.

unstable/wlr-output-manager-v1.xml
```diff
@@ -0,0 +1,350 @@
...
+    <description summary="manage output device properties">
+      Allows the client to change the properties of an output device. The
+      client binds the output manager globally, then creates a configuration
+      object.
+    </description>
+    <request name="create_configuration">
+      <description summary="create a new output configuration object">
+        Create a new output configuration object.
+      </description>
+      <arg name="id" type="new_id" interface="zwlr_output_device_v1"/>
```
emersion: `zwlr_output_configuration_v1`

unstable/wlr-output-manager-v1.xml
```diff
@@ -0,0 +1,350 @@
...
+      <description summary="destroy the output manager">
+        Using this request a client can tell the server that it is not going
+        to use the manager object anymore. Any objects created by the manager
+        will remain valid, until their appropriate destroy request has been
+        called.
+      </description>
+    </request>
+  </interface>
+
+  <interface name="zwlr_output_configuration_v1" version="1">
```
emersion: How do you set the scale?

emersion: No, I mean there's no `set_scale` request in this interface.

Ongy: > Added size. Similar to xdg-output's logical_size event. Allows for fractional scaling.

It is computable from this (provided I understood the intended usecase correctly)
But I wonder if it's a good idea to have this in the long run

emersion: This allows to read the scale, not to set it.

unstable/wlr-output-manager-v1.xml
```diff
@@ -0,0 +1,350 @@
...
+        and whenever the size of the output device in the global compositor
+        space changes.
+      </description>
+      <arg name="width" type="int"
+        summary="width in global compositor space"/>
+      <arg name="height" type="int"
+        summary="height in global compositor space"/>
+    </event>
+
+    <event name="position">
```
emersion: I wonder if it makes sense to have these events at all. These events only make sense if the output is enabled. In this case this data can be obtained from wl_output and xdg-output as normal.

The only added value provided by output_device is that the object exists even if the output is disabled. So I wonder if it's necessary to add these at all.

unstable/wlr-output-manager-v1.xml
VincentVanlaer: I'm not against a link between this and wl_output, but iirc there was some discussion on #wayland about different protocols creating the same interfaces and why that was a bad idea. Might be wrong on that. This would also be a server-created object, which is tricky to destroy. (see http://ppaalanen.blogspot.com/2014/07/wayland-protocol-design-object-lifespan.html )

unstable/wlr-output-manager-v1.xml
VincentVanlaer: My bad, must have forgotten to add the set_size event. I don't think we should drop fractional scaling in this protocol, since it is used for configuration instead of rendering. I used this on X on my old laptop to scale from 1600x900 (built in screen resolution) to 1920x1080 for certain external displays when giving presentations.

unstable/wlr-output-manager-v1.xml
```diff
@@ -0,0 +1,350 @@
...
+
+        This event is sent after the client binds to the output_device and
+        whenever the mode of the output changes.
+      </description>
+      <arg name="width" type="int" summary="width of the output buffer"/>
+      <arg name="height" type="int" summary="height of the output buffer"/>
+      <arg name="refresh" type="int" summary="vertical refresh rate in mHz"/>
+    </event>
+    
+    <event name="transformation">
```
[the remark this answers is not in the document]
VincentVanlaer: I think that would be the cleanest approach.

unstable/wlr-output-manager-v1.xml
VincentVanlaer: The only two events that we would be able to drop are transformation and position. The others are different from the wl_output and xdg_output interfaces. It doesn't feel right to to reference only those two.

unstable/wlr-output-manager-v1.xml
emersion: "A link to wl_output" would not mean that this interface would create wl_output objects.

However it is tricky, because output_device objects are published as globals, and we can "link" wl_outputs only after the client has bound to the global…

unstable/wlr-output-manager-v1.xml
emersion: I'm not sure a `set_size` request would work well: what happens if the compositor doesn't support fractional scaling? What happens if the x-axis ratio is different from the y-axis ratio?

VincentVanlaer: The compositor would reject the change in that case.

unstable/wlr-output-manager-v1.xml
emersion: (Same for "size")

unstable/wlr-output-manager-v1.xml
VincentVanlaer: Ah, got it, that would be tricky indeed.

unstable/wlr-output-manager-v1.xml
emersion: (And "active_mode")

VincentVanlaer: size is different from the xdg_output logical_size in the current description.

unstable/wlr-output-manager-v1.xml
VincentVanlaer: The more I think about this, the more I don't like this approach. An alternative solution would be  to send the name that the registry gives to the global. However, as far as I can tell, it's not possible to obtain that in the current api.

unstable/wlr-output-manager-v1.xml
```diff
@@ -0,0 +1,350 @@
...
+
+        The size described by a mode represents the size of the underlying buffer.
+        It does not need to be the same as the output size in the compositor space. 
+      </description>
+      <arg name="width" type="int" summary="width of the output buffer"/>
+      <arg name="height" type="int" summary="height of the output buffer"/>
+      <arg name="refresh" type="int" summary="vertical refresh rate in mHz"/>
+    </event>
+    
+    <event name="active_mode">
```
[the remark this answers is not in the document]
VincentVanlaer: Maybe the compositor could send zero if it doesn't apply?

unstable/wlr-output-manager-v1.xml
Ongy: Should be client specific, because you may hide others, and they are just a counter from what I've seen.

Well, it would be sent with all other events on binding. It can't be linked to it like xdg-output is either way. It's just an "attribute" the client can track.

agx:
Great this is moving forward! Two things I'm wondering about: where would we put output cloning (one display showing the same as another)? Should this be a part of this protocol?

How would I map a {wl,xdg}_output to an output_device ?

VincentVanlaer:
> Great this is moving forward! Two things I'm wondering about: where would we put output cloning (one display showing the same as another)? Should this be a part of this protocol?

I think that output cloning is to complex for the first version, see https://github.com/swaywm/sway/issues/1666 . However, it would make sense for compositors to check whether `size` and `position` are the same, and clone in that case.

> How would I map a {wl,xdg}_output to an output_device ?

It's being discussed in https://github.com/swaywm/wlr-protocols/pull/23#discussion_r202444038 , but gihub is being annoying and has marked everything outdated.

unstable/wlr-output-manager-v1.xml
VincentVanlaer: Looking at the libwayland code, the names are indeed a counter. However, there's currently no way to obtain the name of a global, except guessing it.

unstable/wlr-output-manager-v1.xml
VincentVanlaer: hmm, if xdg_output.logical_size would be exactly the same, I'd rather have them separate too. That would solve https://github.com/swaywm/wlr-protocols/pull/23#discussion_r202509810 .

VincentVanlaer:
Should have addressed all comments except for
* https://github.com/swaywm/wlr-protocols/pull/23#discussion_r202444038
* https://github.com/swaywm/wlr-protocols/pull/23#discussion_r202512759
* https://github.com/swaywm/wlr-protocols/pull/23#discussion_r202509810

unstable/wlr-output-management-unstable-v1.xml
```diff
@@ -0,0 +1,361 @@
...
+      <description summary="advertise supported output modes">
+        The supported_mode event describes an available mode for the output.
+        These modes are advertised by the output device as being supported and
+        should be preferred over any other, custom, mode.
+
+        When the client binds to the output_device object, the server sends this
+        event once for every supported mode.
+
+        If the concept of a specific refresh rate does not apply to the mode,
+        the refresh argument shpuld be zero. An example of this is a monitor
```
emersion: Typo

unstable/wlr-output-management-unstable-v1.xml
```diff
@@ -0,0 +1,361 @@
...
+        This event is sent after the client binds to the output_device
+        and whenever the position of the output changes in the compositor space.
+      </description>
+      <arg name="x" type="int"
+        summary="x position within the global compositor space"/>
+      <arg name="y" type="int"
+        summary="y position within the global compositor space"/>
+    </event>
+    
+    <enum name="enablement">
```
emersion: This is not used anymore

SirCmpwn:
I would prefer to refactor this into a design similar to xdg-output, where the client explicitly maps wl_outputs to zwlr_output_device_v1, for consistnecy and so that we don't have to duplicate the information provided by xdg_output.

VincentVanlaer:
@emersion Unless I missed something, it just stops at v4 without any comments. The discussion might be interesting, I'll give it a read.

@SirCmpwn Aye, I agree that duplication is not optimal, but we still want to represent disabled outputs. Should we have a global for each disabled output? So one either has a wl_output or a wl_disabled_output for each output that is connected. It makes the configuration interface messy though since it needs to accept either type of output.

SirCmpwn:
Right, I forgot about disabled outputs. Bleh.

emersion:
One issue this protocol doesn't address is frame-perfect output configuration on hotplug.

When the logic to configure outputs is in the compositor, it's easy to apply it on hotplug and get the desired configuration on first modeset. When the logic is in a client like this, we don't know if the client will attempt to change it on hotplug so there will be two modesets: one done by the compositor after hotplug, and another one done by the client to apply the new configuration.

For instance if I have configured an output to be disabled in my client config, plugging it will first enable it and then disable it. Same if I've configured a different mode.

Maybe this could be solved with a `configure` event similar to xdg-shell.

Ongy:
Calling this "frame perfect" is funny with the time some outputs take to modeset :)

The interesting challenge here is to provide a mechanism that will work without an application and can be driven with one modeset from the an application.

One simple way would be to mandate compositors don't automatically change related settings while any application has an object of this protocol.
Though I'd think a kind of capability field that tells the compositor what the client wants to do (e.g. re-position outputs when another is removed, manage freshly hotpluged etc.) so we don't have to have all features in every client.

emersion:
Additional prior art: GNOME https://wiki.gnome.org/Initiatives/Wayland/Gaps/DisplayConfig

SirCmpwn:
>Calling this "frame perfect" is funny with the time some outputs take to modeset :)

We should design this protocol under the assumption that 10 years from now it'll be used to modeset outputs which can change modes within a single frame

agx:
> One issue this protocol doesn't address is frame-perfect output configuration on hotplug.
> 
> When the logic to configure outputs is in the compositor, it's easy to apply it on hotplug and get the desired configuration on first modeset. When the logic is in a client like this, we don't know if the client will attempt to change it on hotplug so there will be two modesets: one done by the compositor after hotplug, and another one done by the client to apply the new configuration.
> 
> For instance if I have configured an output to be disabled in my client config, plugging it will first enable it and then disable it. Same if I've configured a different mode.
> 
> Maybe this could be solved with a `configure` event similar to xdg-shell.

For hotplugged outputs (when the compositor is already running) we can keep them off and then only apply the configuration once the compositor sees the "apply" request, no? It's more cumbersome when the compositor starts up and applies a default configuraton and the client isn't happy with that which would introduce screen flickering.

emersion:
>For hotplugged outputs (when the compositor is already running) we can keep them off and then only apply the configuration once the compositor sees the "apply" request, no?

Yes, but there are issues with output changes not seen as atomic. So we should probably have an event saying "I've sent all newly plugged outputs and removed all newly unplugged outputs, now please send a configuration" and an event saying "you asked me to configure all outputs, please apply this state atomically". This will help in cases where there are less CRTCs than connected outputs.

emersion:
Here is a new proposal: https://github.com/swaywm/wlr-protocols/pull/38

PR #24, opened 2018-07-18
Add gamma control

emersion:
This uses file descriptors instead of arrays to fix https://github.com/swaywm/wlroots/issues/1135.

It also adds descriptions and clearer semantics. `reset_gamma` is removed in favour of using the `destroy` request, and a `failed` event is added.

As always, comments are welcome. Do we want anything else in this protocol?

### DIFF
--- a/unstable/wlr-gamma-control-unstable-v1.xml
+++ b/unstable/wlr-gamma-control-unstable-v1.xml
@@ -72,7 +72,7 @@
       tables. At any time the compositor can send a failed event indicating that
       this object is no longer valid.
 
-      There must always be at most one gamma control object per output, which
+      There can only be at most one gamma control object per output, which
       has exclusive access to this particular output. When the gamma control
       object is destroyed, the gamma table is restored to its original value.
     </description>
--- a/unstable/wlr-gamma-control-unstable-v1.xml
+++ b/unstable/wlr-gamma-control-unstable-v1.xml
@@ -1,0 +1,126 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<protocol name="wlr_gamma_control_unstable_v1">
+  <copyright>
+    Copyright © 2015 Giulio camuffo
+    Copyright © 2018 Simon Ser
+
+    Permission to use, copy, modify, distribute, and sell this
+    software and its documentation for any purpose is hereby granted
+    without fee, provided that the above copyright notice appear in
+    all copies and that both that copyright notice and this permission
+    notice appear in supporting documentation, and that the name of
+    the copyright holders not be used in advertising or publicity
+    pertaining to distribution of the software without specific,
+    written prior permission.  The copyright holders make no
+    representations about the suitability of this software for any
+    purpose.  It is provided "as is" without express or implied
+    warranty.
+
+    THE COPYRIGHT HOLDERS DISCLAIM ALL WARRANTIES WITH REGARD TO THIS
+    SOFTWARE, INCLUDING ALL IMPLIED WARRANTIES OF MERCHANTABILITY AND
+    FITNESS, IN NO EVENT SHALL THE COPYRIGHT HOLDERS BE LIABLE FOR ANY
+    SPECIAL, INDIRECT OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN
+    AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION,
+    ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF
+    THIS SOFTWARE.
+  </copyright>
+
+  <description summary="manage gamma tables of outputs">
+    This protocol allows a privileged client to set the gamma tables for
+    outputs.
+
+    Warning! The protocol described in this file is experimental and
+    backward incompatible changes may be made. Backward compatible changes
+    may be added together with the corresponding interface version bump.
+    Backward incompatible changes are done by bumping the version number in
+    the protocol and interface names and resetting the interface version.
+    Once the protocol is to be declared stable, the 'z' prefix and the
+    version number in the protocol and interface names are removed and the
+    interface version number is reset.
+  </description>
+
+  <interface name="zwlr_gamma_control_manager_v1" version="1">
+    <description summary="manager to create per-output gamma controls">
+      This interface is a manager that allows creating per-output gamma
+      controls.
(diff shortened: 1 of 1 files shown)


emersion:
>I think you should be able to disconnect and leave gamma as it was, no reason to leave a process hanging around. Consider that gamma ramps might be used for color tuning as well.

I really don't think it's a good idea. Setting an invalid gamma table can make your output unusable. In general Wayland clients cannot set permanent state and I think this is a good thing.

Color management will be more involved, so it'll most likely require a separate protocol (see https://lists.freedesktop.org/archives/wayland-devel/2016-November/031729.html).

martinetd:
fwiw I run redshift in one-shot mode from key bindings, having it carrying on running from there is definitely not impossible but smooth transition from one step to the next might suck a bit... (would need to start the next redshift, wait until it applied its ramp, then kill the previous one; otherwise killing the previous one would reset the gamma ramp once and flicker the screen)

However I do agree that in general not being able to leave a broken state behind is a good thing, so I guess it's just a matter of fixing the usage pattern e.g. giving redshift a proper remote-control protocol instead.

SirCmpwn:
Can't they just leave it broken by fucking up the gamma and then not disconnecting?

SirCmpwn:
How are you going to debug it, kill it, etc if your screens are hosed?

martinetd:
>> giving redshift a proper remote-control protocol instead
>
> Yeah this is the intended behaviour.

I meant a remote-control protocol to control redshift, not to adjust the gamma ramp. There's currently no way to manually adjust the screen temperature other than one-shot mode, which exits immediately after changing the temperature. I wonder why people figure time and location are any good to decide how red the screen should be in the days of artificial light -_-
But anyway, I'm just ranting at this point, I'll deal with it when this comes out however that turns out to be.

agx:
LGTM. I'd also go for not resetting the gamma ramp when he client disconnects, since (as you wrote) it's for privileged clients.

Ongy:
> How are you going to debug it, kill it, etc if your screens are hosed?

ssh in? At least you got an offender, *and* can fix it without digging out a potentially new client that then just needs to do some repairs

martinetd:
> This mechanism also ensures that there aren't two clients setting gamma ramps at the same time.

Yeah that actually makes it so I won't be able to hack in two "set-and-daemonize" one-shot instance of redshifts to set next gradiant of gamma before killing the previous one, I almost complained when I realized that I would actually have to give this a proper control implementation instead :P

tbh I think this would be fine with a "latest who gives an order wins" design like it does on X11, would just need to be careful on reset to reset the previous last state on exit (and not reset anything if exiting client wasn't the last one), but both are valid designs so I don't really care either way.

>> I meant a remote-control protocol to control redshift
>
> Yeah, that's what I understood. I think redshift should allow this (doesn't it already allow this via its traybar icon menu?).

My bad, I had read your answer with "this" being "this new protocol".

You just made me install redshift-gtk to try the tray icon, it doesn't seem to have any gimmick other than disabling redshift altogether temporarily, but I might be missing out. It does this with SIGUSR1 == toggle, so that's not really exactly extensible; the report info given in the widget come from stdout of the redshift daemon.
So no, will have to do something from scratch.

FWIW though, Quartz (OSX) has the same behaviour of resetting the gamma ramp when redshift exits, so you can pretend we're being cool by doing like Apple (sorry, couldn't resist)
Anyway, it's not overly difficult to make manual mode read from stdin and adjust everytime it gets a line and use a pipe or something, it's just one more thing to do -- not changing grumpy people's habits isn't a valid reason to not do things you think are right.

unstable/wlr-gamma-control-unstable-v1.xml
```diff
@@ -0,0 +1,126 @@
...
+  <interface name="zwlr_gamma_control_v1" version="1">
+    <description summary="adjust gamma tables for an output">
+      This interface allows a client to adjust gamma tables for a particular
+      output.
+
+      The client will receive the gamma size, and will then be able to set gamma
+      tables. At any time the compositor can send a failed event indicating that
+      this object is no longer valid.
+
+      There must always be at most one gamma control object per output, which
```
minus7: Maybe define it here in the protocol that it's not possible to obtains a second object for the same output (so it's not implementation-defined whether the second one will invalidate the first one or not). Or the other way around if that makes more sense (I think it doesn't)

unstable/wlr-gamma-control-unstable-v1.xml
```diff
@@ -0,0 +1,126 @@
...
+    </request>
+  </interface>
+
+  <interface name="zwlr_gamma_control_v1" version="1">
+    <description summary="adjust gamma tables for an output">
+      This interface allows a client to adjust gamma tables for a particular
+      output.
+
+      The client will receive the gamma size, and will then be able to set gamma
+      tables. At any time the compositor can send a failed event indicating that
```
minus7: Is that the `invalid_gamma` error?

unstable/wlr-gamma-control-unstable-v1.xml
minus7: Oh, I see, the decision what to do *should* lie with the compositor. Maybe change the wording to "There can only be.." so it's clear that it's not the user's duty to check if they're the only one.

unstable/wlr-gamma-control-unstable-v1.xml
emersion: No. Protocol errors are reserved for situations in which the client does something wrong. Protocol errors are suitable for a client passing a gamma with an invalid size, but are unsuitable for signaling that an output doesn't support gamma control.

unstable/wlr-gamma-control-unstable-v1.xml
emersion: Right.

unstable/wlr-gamma-control-unstable-v1.xml
```diff
@@ -0,0 +1,126 @@
...
+      this object is no longer valid.
+
+      There can only be at most one gamma control object per output, which
+      has exclusive access to this particular output. When the gamma control
+      object is destroyed, the gamma table is restored to its original value.
+    </description>
```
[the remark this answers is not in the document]
emersion: - I'm not sure what the upside is, since you'll need to hold the gamma_control_manager instead of the gamma_control?
- Clients are supposed to keep the gamma_control around to keep a "lock" on the output and prevent other clients from controlling the same output. How would this mechanism work with your proposal?
- A common Wayland idiom is to create the manager, create the control, and immediately destroy the manager. This is done with `wl_region` for instance (clients create a region, assign it to a surface, and then immediately destroy it).

Issue #10, opened 2018-04-11
Virtual desktop protocol


acrisci:
Nearly every window manager has a concept of a "virtual desktop" in their operation.

i3 has the concept of "workspaces" and Awesome has "tags" which are nearly identical concepts. Each output has a currently active workspace that has its own set of visible windows, and inactive workspaces that are hidden. Many workspaces can exist on an output and each output has its own set of workspaces. Outputs can switch workspaces independently of each other.

Xfce and Gnome have "virtual desktops". There is a single set of virtual desktops for the whole screen. This nomenclature comes from the `_NET_DESKTOP_*` [EWMH](https://standards.freedesktop.org/wm-spec/wm-spec-latest.html#idm140200472728352) root window properties and client requests.

I would like to support both paradigms if possible.

The goal of this protocol is to support out of process bars that can be run on any window manager. This was never possible on X11 because EWMH only supported the "virtual desktop" paradigm.

Required of the protocol:

*List all of the workspaces for each output.*

This is to display the workspaces on the bar and their active status.

*A way to request a change in the active workspace or a create a new workspace*

A bar might request this when the user interacts with the bar like clicking a button. The compositor should be able to deny the request.

*Workspaces on multiple outputs at once*

This is to support the traditional virtual desktop paradigm. This might require desktop status to be double buffered to support changing workspaces on several outputs at once.

The first target compositors are sway and way-cooler.

Let me know if you have concerns. If you like the idea, I'll write some XML.

SirCmpwn:
My gut reaction to this protocol is to be cautiously optimistic, emphasis on cautiously. I think a good protocol for this may be hard to write.

>Workspaces on multiple outputs at once

I'm not sure that it makes sense for the sake of GNOME-like paradigms to make workspaces always a child of an output. Even sway may eventually provide config knobs to change its behavior in this way (you can already hack it together). We should make workspaces a standalone concept that can optionally have any number of outputs associated with them.

>A way to request a change in the active workspace or a create a new workspace

Change, yes. Create, not so sure.

We should also consider the possibility of one output having several visible/active outputs. Swaybar also needs to be able to distinguish between hidden, visible, focused, and focus-inactive workspaces.

acrisci:
> I think a good protocol for this may be hard to write.

Absolutely. But the benefit would be great. I think we all want to have modular desktops where we can choose the components that work best for us. We never had a thriving bar ecosystem on X11 so I'm motivated to give it a shot.

[...]

PR #26, opened 2018-08-11
add wlr-foreign-toplevel-management protocol to enumerate and control desktop surfaces

### LINKED ISSUE #14: Add protocol for enumerating shell surfaces and their locations
Locations respect to xdg-output, and the protocol should assume a surface could be rendered in several places.

ammen99:
I'm opening this PR to discuss what we need to have in this protocol. I'm not very familiar with writing wayland protocols "the proper way", so any recommendations are very welcome :)

Things I'm not sure about:

1. Thumbnails - some prior art like wayland-wall required the compositor draw the view in a specified place on the screen. However, I think this is better solved if the app could create screenshots of individual windows.

2. Minimizing. When the compositor supports more advanced window minimize animations, it usually needs a place on the screen to minimize towards (think of the magic lamp effect in compiz).

I think this fixes #14, but I'm not sure whether consensus was reached if we should include window positions or not.

### DIFF
--- a/unstable/wlr-foreign-toplevel-management-unstable-v1.xml
+++ b/unstable/wlr-foreign-toplevel-management-unstable-v1.xml
@@ -1,0 +1,235 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<protocol name="wlr_foreign_toplevel_management_unstable_v1">
+  <copyright>
+    Copyright © 2018 Ilia Bozhinov
+
+    Permission to use, copy, modify, distribute, and sell this
+    software and its documentation for any purpose is hereby granted
+    without fee, provided that the above copyright notice appear in
+    all copies and that both that copyright notice and this permission
+    notice appear in supporting documentation, and that the name of
+    the copyright holders not be used in advertising or publicity
+    pertaining to distribution of the software without specific,
+    written prior permission.  The copyright holders make no
+    representations about the suitability of this software for any
+    purpose.  It is provided "as is" without express or implied
+    warranty.
+
+    THE COPYRIGHT HOLDERS DISCLAIM ALL WARRANTIES WITH REGARD TO THIS
+    SOFTWARE, INCLUDING ALL IMPLIED WARRANTIES OF MERCHANTABILITY AND
+    FITNESS, IN NO EVENT SHALL THE COPYRIGHT HOLDERS BE LIABLE FOR ANY
+    SPECIAL, INDIRECT OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN
+    AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION,
+    ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF
+    THIS SOFTWARE.
+  </copyright>
+
+  <interface name="zwlr_foreign_toplevel_manager_v1" version="1">
+    <description summary="list and control opened apps">
+      The purpose of this protocol is to enable the creation of taskbars
+      and docks by providing them with a list of opened applications and
+      letting them request certain actions on them, like maximizing, etc.
+
+      After a client binds the zwlr_foreign_toplevel_manager_v1, each opened
+      toplevel window will be sent via the toplevel event
+    </description>
+
+    <event name="toplevel">
+      <description summary="a toplevel has been created">
+        This event is emitted whenever a new toplevel window is created. It
+        is emitted for all toplevels, regardless of the app that has created
+        them.
+
+        All initial details of the toplevel(title, app_id, states, etc.) will
+        be sent immediately after this event via the corresponding events in
+        zwlr_foreign_toplevel_handle_v1.
+      </description>
+      <arg name="toplevel" type="new_id" interface="zwlr_foreign_toplevel_handle_v1"/>
+    </event>
+
+    <request name="stop">
+      <description summary="stop sending events">
+        Indicates the client no longer wishes to receive events for new toplevels.
+        However the compositor may emit further toplevel_created events, until
+        the finished event is emitted.
+
+        The client must not send any more requests after this one.
+      </description>
+    </request>
+
+    <event name="finished">
+      <description summary="the compositor has finished with the toplevel manager">
+        This event indicates that the compositor is done sending events to the
+        zwlr_foreign_toplevel_manager_v1. The server will destroy the object
+        immediately after sending this request, so it will become invalid and
+        the client should free any resources associated with it.
+      </description>
+    </event>
+  </interface>
+
+  <interface name="zwlr_foreign_toplevel_handle_v1" version="1">
+    <description summary="an opened toplevel">
+      A zwlr_foreign_toplevel_handle_v1 object represents an opened toplevel
+      window. Each app may have multiple opened toplevels.
+
+      Each toplevel has a list of outputs it is visible on, conveyed to the
+      client with the output_enter and output_leave events.
+    </description>
+
+    <event name="title">
+      <description summary="title change">
+        This event is emitted whenever the title of the toplevel changes.
+      </description>
+      <arg name="title" type="string"/>
+    </event>
+
+    <event name="app_id">
+      <description summary="app-id change">
+        This event is emitted whenever the app-id of the toplevel changes.
+      </description>
+      <arg name="app_id" type="string"/>
+    </event>
+
+    <event name="output_enter">
+      <description summary="toplevel entered an output">
+        This event is emitted whenever the toplevel becomes visible on
+        the given output. A toplevel may be visible on multiple outputs.
+      </description>
+      <arg name="output" type="object" interface="wl_output"/>
+    </event>
+
+    <event name="output_leave">
+      <description summary="toplevel left an output">
+        This event is emitted whenever the toplevel stops being visible on
+        the given output. It is guaranteed that an entered-output event
+        with the same output has been emitted before this event.
+      </description>
+      <arg name="output" type="object" interface="wl_output"/>
+    </event>
+
+    <request name="set_maximized">
+      <description summary="requests that the toplevel be maximized">
+        Requests that the toplevel be maximized. If the maximized state actually
+        changes, this will be indicated by the state event.
+      </description>
+    </request>
+
+    <request name="unset_maximized">
+      <description summary="requests that the toplevel be unmaximized">
+        Requests that the toplevel be unmaximized. If the maximized state actually
+        changes, this will be indicated by the state event.
+      </description>
+    </request>
+
+    <request name="set_minimized">
+      <description summary="requests that the toplevel be minimized">
+        Requests that the toplevel be minimized. If the minimized state actually
+        changes, this will be indicated by the state event.
+      </description>
+    </request>
+
+    <request name="unset_minimized">
+      <description summary="requests that the toplevel be unminimized">
+        Requests that the toplevel be unminimized. If the minimized state actually
+        changes, this will be indicated by the state event.
+      </description>
+    </request>
+
+    <request name="activate">
+      <description summary="activate the toplevel">
+        Request that this toplevel be activated on the given seat.
+        There is no guarantee the toplevel will be actually activated.
+      </description>
+      <arg name="seat" type="object" interface="wl_seat"/>
+    </request>
+
+    <enum name="state">
+      <description summary="types of states on the toplevel">
+        The different states that a toplevel can have. These have the same meaning
+        as the states with the same names defined in xdg-toplevel
+      </description>
+
+      <entry name="maximized" value="0" summary="the toplevel is maximized"/>
+      <entry name="minimized" value="1" summary="the toplevel is minimized"/>
+      <entry name="activated" value="2" summary="the toplevel is active"/>
+    </enum>
+
+    <event name="state">
+      <description summary="the toplevel state changed">
+        This event is emitted immediately after the zlw_foreign_toplevel_handle_v1
+        is created and each time the toplevel state changes, either because of a
+        compositor action or because of a request in this protocol.
+      </description>
+
+      <arg name="state" type="array"/>
+    </event>
+
+    <event name="done">
+      <description summary="all information about the toplevel has been sent">
+        This event is sent after all changes in the toplevel state have been
+        sent.
+
+        This allows changes to the zwlr_foreign_toplevel_handle_v1 properties
+        to be seen as atomic, even if they happen via multiple events.
+      </description>
+    </event>
+
+    <request name="close">
+      <description summary="request that the toplevel be closed">
+        Send a request to the toplevel to close itself. The compositor would
+        typically use a shell-specific method to carry out this request, for
+        example by sending the xdg_toplevel.close event. However, this gives
+        no guarantees the toplevel will actually be destroyed. If and when
+        this happens, the zwlr_foreign_toplevel_handle_v1.closed event will
+        be emitted.
+      </description>
+    </request>
+
+    <request name="set_rectangle">
+      <description summary="the rectangle which represents the toplevel">
+        The rectangle of the surface specified in this request corresponds to
+        the place where the app using this protocol represents the given toplevel.
+        It can be used by the compositor as a hint for some operations, e.g
+        minimizing. The client is however not required to set this, in which
+        case the compositor is free to decide some default value.
+
+        If the client specifies more than one rectangle, only the last one is
+        considered.
+
+        The dimensions are given in surface-local coordinates.
+        Setting width=height=0 removes the already-set rectangle.
+      </description>
+
+      <arg name="surface" type="object" interface="wl_surface"/>
+      <arg name="x" type="int"/>
+      <arg name="y" type="int"/>
+      <arg name="width" type="int"/>
+      <arg name="height" type="int"/>
+    </request>
+
+    <enum name="error">
+      <entry name="invalid_rectangle" value="0"
+        summary="the provided rectangle is invalid"/>
+    </enum>
+
+    <event name="closed">
+      <description summary="this toplevel has been destroyed">
+        This event means the toplevel has been destroyed. It is guaranteed there
+        won't be any more events for this zwlr_foreign_toplevel_handle_v1. The
+        toplevel itself becomes inert so any requests will be ignored except the
+        destroy request.
+      </description>
+    </event>
+
+    <request name="destroy" type="destructor">
+      <description summary="destroy the zwlr_foreign_toplevel_handle_v1 object">
+        Destroys the zwlr_foreign_toplevel_handle_v1 object.
+
+        This request should be called either when the client does not want to
+        use the toplevel anymore or after the closed event to finalize the
+        destruction of the object.
+      </description>
+    </request>
+  </interface>
+</protocol>


unstable/wlr-window-list.xml
```diff
@@ -0,0 +1,144 @@
...
+
+  <interface name="zwlr_window_v1" version="1">
+    <description summary="an opened window">
+      A zwlr_window_v1 object represents an opened window.
+      Each app may have multiple opened windows.
+    </description>
+
+    <event name="title">
+      <description summary="iitle change">
+        This event is emitted whenever the title of the window changes
```
agx: title?

unstable/wlr-window-list.xml
```diff
@@ -0,0 +1,144 @@
...
+    SPECIAL, INDIRECT OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN
+    AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION,
+    ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF
+    THIS SOFTWARE.
+  </copyright>
+
+  <interface name="zwlr_window_manager_v1" version="1">
+    <description summary="list and control opened apps">
+      The purpose of this protocol is to enable the creation of taskbars
```
agx: Is this also about controlling or just about listing?

unstable/wlr-window-list.xml
```diff
@@ -0,0 +1,144 @@
...
+
+      Each created window has a list of outputs it is visible on.
+    </description>
+
+    <event name="window_created">
+      <description summary="a window has been created">
+        This event is emitted whenever a new toplevel window is created.
+      </description>
+      <arg name="window" type="new_id" interface="zwlr_window_v1"/>
+    </event>
```
agx: Is the list of current windows supposed to be sent when the client binds to zwlr_window_manager_v1? If so the protocol should specify it.

unstable/wlr-window-list.xml
```diff
@@ -0,0 +1,144 @@
...
+
+    <event name="app_id">
+      <description summary="app-id change">
+        This event is emitted whenever the app-id of the window changes
+        or after it is created (the initial title)
+      </description>
+      <arg name="app_id" type="string"/>
+    </event>
+
+    <event name="enter_output">
```
agx: Should that be "entered_output" to match the below "maximized"?

unstable/wlr-window-list.xml
ammen99: also controlling, there are focus/maximize/minimize/close requests

unstable/wlr-window-list.xml
ammen99: yes, it is so, will do it.

unstable/wlr-window-list.xml
ammen99: no, it is more like the `wl_surface.enter` event

ammen99:
@agx we might want to indicate that the window is xwayland and give access to the window ID so that the client can get the app icon. Aside from that I don't see any reason to expose the "type" of the window. 
Something I wondered, when I wrote this I was focused on "regular" windows. Is it clear panels/backgrounds are excluded?

emersion:
Consider mirroring xdg-toplevel and using xdg-toplevel's vocabulary.

> Is it clear panels/backgrounds are excluded?

IMHO yes.

>Thumbnails - some prior art like wayland-wall required the compositor draw the view in a specified place on the screen. However, I think this is better solved if the app could create screenshots of individual windows.

I don't agree. I think adding a request creating `wl_buffer`s that mirror other views and can be attached to surfaces is a better approach.

>Minimizing. When the compositor supports more advanced window minimize animations, it usually needs a place on the screen to minimize towards (think of the magic lamp effect in compiz).

If included, this should be optional.

unstable/wlr-window-list.xml
```diff
@@ -0,0 +1,148 @@
...
+      <description summary="a window has been created">
+        This event is emitted whenever a new toplevel window is created.
+      </description>
+      <arg name="window" type="new_id" interface="zwlr_window_v1"/>
+    </event>
+
+    <request name="destroy" type="destructor"/>
+  </interface>
+
+  <interface name="zwlr_window_v1" version="1">
```
SirCmpwn: I think this language is a bit too generic. Imagine this eventually becomes stable and we're looking at a time named `xdg_window`. I think maybe `window_handle` or `foreign_toplevel` would be better.

unstable/wlr-window-list.xml
SirCmpwn: Turning that question around: should `maximized` be called `maximize`?

SirCmpwn: Also, I'd prefer `output_leave` and `output_enter` over the reverse

emersion:
>Okay, so that includes splitting the maximize/minimize request into set_maximized and unset_maximized, and sending the states as an array? (if I got you correctly)

I'd also rename "window" → "toplevel". Windows can be popups (and your clipboard too! and your screen too!) in X11, sticking to the xdg-shell vocabulary makes things clearer.

unstable/wlr-window-list.xml
```diff
@@ -0,0 +1,148 @@
...
+      <arg name="output" type="object" interface="wl_output"/>
+    </event>
+
+    <event name="virtual_desktop">
+      <description summary="window was moved to a virtual desktop">
+        This event is emitted with the initial virtual desktop the window
+        is on and whenever it is moved to a different virtual desktop later.
+      </description>
+      <arg name="desktop" type="string"/>
+    </event>
```
SirCmpwn: This probably belongs in a separate protocol. It's a subtle issue.

https://github.com/swaywm/wlr-protocols/issues/10

unstable/wlr-window-list.xml
```diff
@@ -0,0 +1,148 @@
...
+      </description>
+      <arg name="minimized" type="uint"/>
+    </request>
+
+    <request name="focus">
+      <description summary="request focus">
+        Request that this window is focused. However, there is no guarantee
+        the window will be actually focused.
+      </description>
+    </request>
```
SirCmpwn: Should we call this "activate"? Also, there's no event which tells you if the window is focused or not.

ammen99: hmm, yes, this will better fit with xdg-shell wording

unstable/wlr-window-list.xml
ammen99: okay, will remove it for now

ammen99:
I updated the protocol to use `foreign_toplevel` and also addressed the other comments. I also added the set_rectangle request, which is used to specify where on the screen the toplevel is represented

agx:
Since activation is part of the protocol should raising the window to the top of the stack also be or is this considered to be compositor specific (like raise when activated if necessary)?

ammen99:
> For e.g. terminating applications it would be good to have the pid of the client that owns the surface. Should there be a request for that too?

IMO force-kill belongs to the compositor, not the dock/taskbar/whatever

agx:
> IMO force-kill belongs to the compositor, not the dock/taskbar/whatever

Imagine a touch device where the "taskbar" is fullscreen and lists all running applications to switch to them. Kill/Forcekilling fits in there nicely. Same holds for a HUD on the desktop. An alternative would be to have "kill client matching toplevel" request so we don't need to expose the pid.

ammen99:
@agx I was thinking of desktop usage, where you usually get "force-kill or wait" dialogue. But indeed on a smartphone it would make more sense to just kill it. But isn't the task switcher built-in into the compositor in this case?

Ongy:
PID doesn't necessarily make sense.
And when the socket was created with `socketpair()` and passed to the client, I don't think it's necessarily possible to get it.

So the way to go should be over the compositor that simply disconnects the client as worst case measure (or tries to crawl /proc/ to get the pid, but that's not very nice either)

SirCmpwn:
I think the compositor can use xdg-shell ping shortly after it's been asked to close a surface, and if the client doesn't respond it can show its own UI for force killing the application.

agx:
Well that implies that the compositor wants to do UI as well (with all implications like pulling in a UI toolkit etc). I don't think the protocol should mandate that. Since the compositor can't be restarted I'd rather have these things outside of it. That doesn't mean that it has to be in *this* protocol but since maximize and minimize is in there too it fits.

SirCmpwn:
I think it's giving the client too much power. A compositor that doesn't want to make the UI for that could just force kill clients when a second close request is made after enough time without responding to a ping. It could also change the title to "[Unresponsive] $title" or something, too.

ammen99:
@agx if you want force-kill, then why not bake this into the compositor? I think desktop compositors will implement such a popup in some way anyway, and for touch-based devices, just kill after a short timeout by default.

agx:
Having the compositor run the same UI toolkit than the shell makes little sense (and since it's not restartable I want as little baggage in the compositor as possible). Doing that in process is just not necessary (neither on the desktop nor on mobile devices).

I also don't buy the "power" argument since clients allowed to certain protocols have lots of power already (like faking arbitrary appliacations by using screen shot and layer surface). As I said it doesn' t have to go into this protocol, I can always add a separate one on top.

SirCmpwn:
Alright, separate protocol seems fine.

agx:
Rereading this: the last comments only focus on force killing but I'm also talking about terminating application (SIGTERM) e.g. the shell displays the toplevels and you swipe up to close the application or like gnome-shell displays a little "x" in the top-right corner of the overview that can be clicked to close the application. 
(Just for completeness so it doesn't get lost when I pick it up again)

emersion:
>terminating application (SIGTERM) e.g. the shell displays the toplevels and you swipe up to close the application or like gnome-shell displays a little "x" in the top-right corner of the overview that can be clicked to close the application.

This is done via a "close" event in xdg-shell (and a similar mechanism on X11 when the client supports it). This is not done by sending a signal to the process.

agx:
@emersion good point. My point was basically that it's not only about "force-kill" but a also about application windows. Using the means of the underlying protocol (e.g. xdg-shell) is certainly better than plain unix signals.

So maybe plain "close" fits in here while force-kill does not?

ammen99:
But @agx is right, such a request (like xdg-shell close) belongs to this protocol, and I thought I had included it. Maybe I deleted it accidentally, will fix that.

ammen99:
Is the newly added close-request satisfactory?

agx:
@ammen99 looks good to me!

SirCmpwn:
LGTM

unstable/wlr-foreign-toplevel-list.xml
```diff
@@ -0,0 +1,190 @@
...
+    <event name="toplevel_created">
+      <description summary="a toplevel has been created">
+        This event is emitted whenever a new toplevel window is created. It
+        is emitted for all toplevels, regardless of the app that has created
+        them.
+      </description>
+      <arg name="toplevel" type="new_id" interface="zwlr_foreign_toplevel_v1"/>
+    </event>
+
+    <request name="destroy" type="destructor"/>
```
emersion: This should have a description explaining what happens to objects created via `toplevel_created`.

It's tricky to design an interface that doesn't leak.

ammen99: do we need to explain that after I changed the proto to use client-generated objects?

emersion: Yes.

ammen99:
I do hope the protocol is in a better shape now, thank you all for the patience.

ammen99:
Yet another change, inspired by a discussion in #wayland:

> use server-side created resources, add a request on the factory to ask the compositor to stop sending resources in events

unstable/wlr-foreign-toplevel-list.xml
```diff
@@ -0,0 +1,217 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<protocol name="wlr_foreign_toplevel_list_v1">
```
emersion: The name should be `wlr_foreign_toplevel_list_unstable_v1`

unstable/wlr-foreign-toplevel-list.xml
```diff
@@ -0,0 +1,217 @@
...
+    <description summary="list and control opened apps">
+      The purpose of this protocol is to enable the creation of taskbars
+      and docks by providing them with a list of opened applications and
+      letting them request certain actions on them, like maximizing, etc.
+
+      After a client binds the zwlr_foreign_toplevel_manager_v1, each opened
+      toplevel window will be sent via the toplevel_created signal.
+    </description>
+
+    <event name="toplevel_created">
```
emersion: This can just be named `toplevel` to follow the Wayland naming conventions (e.g. `wl_output` has a `scale` event, `wl_data_device` has `data_offer`...)

unstable/wlr-foreign-toplevel-list.xml
```diff
@@ -0,0 +1,217 @@
...
+        zwlr_foreign_toplevel_v1.
+      </description>
+      <arg name="toplevel" type="new_id" interface="zwlr_foreign_toplevel_v1"/>
+    </event>
+
+    <request name="stop">
+      <description summary="stop sending events">
+        Indicates the client no longer wishes to receive events for new toplevels.
+        However the compositor may emit further toplevel_created events, until
+        the finished event is emitted.
```
emersion: Add that the client cannot send more requests after this one

unstable/wlr-foreign-toplevel-list.xml
```diff
@@ -0,0 +1,217 @@
...
+      </description>
+    </request>
+
+    <event name="finished">
+      <description summary="the compositor has finished with the toplevel manager">
+        This event indicates that the compositor is done sending events to the
+        zwlr_foreign_toplevel_manager_v1. The server will destroy the resource
+        immediately after sending this request. The client should still call
+        zwlr_foreign_toplevel_v1_destroy() as a response without sending any
```
emersion: This is not really a protocol concern, and mentioning a function call ties this protocol to libwayland. I'd just say something among the lines of "the object is not valid anymore and any resource associated to the object should be released".

unstable/wlr-foreign-toplevel-list.xml
```diff
@@ -0,0 +1,217 @@
...
+        The different states that a toplevel can have. These have the same meaning
+        as the states with the same names defined in xdg-toplevel
+      </description>
+
+      <entry name="maximized" value="0" summary="the toplevel is maximized"/>
+      <entry name="minimized" value="1" summary="the toplevel is minimized"/>
+      <entry name="activated" value="2" summary="the toplevel is active"/>
+    </enum>
+
+    <event name="state_changed">
```
emersion: This event can just be named `state`.

I wonder if we want a way to atomically commit state (like the `commit` event for `wl_surface`, or the `configure` event for `xdg_toplevel`). This might be useful if someone wants to design new protocols on top of this one to retrieve more properties.

unstable/wlr-foreign-toplevel-list.xml
```diff
@@ -0,0 +1,217 @@
...
+      <arg name="width" type="int"/>
+      <arg name="height" type="int"/>
+    </request>
+
+    <event name="destroyed">
+      <description summary="this toplevel has been destroyed">
+        This event means the toplevel has been destroyed. It is guaranteed there
+        won't be any more events for this zwlr_foreign_toplevel_v1. The client
+        should destroy the zwlr_foreign_toplevel_v1 object without sending any
+        further requests.
```
emersion: >without sending any further requests

This is racy. Is this a protocol error?

You might want to just say something like "the object becomes inert and any further request will be ignored" instead.

unstable/wlr-foreign-toplevel-list.xml
ammen99: I guess I should rename the file, too?

unstable/wlr-foreign-toplevel-list.xml
ammen99: okay

unstable/wlr-foreign-toplevel-list.xml
ammen99: I also thought about that, will do.

unstable/wlr-foreign-toplevel-list.xml
ammen99: Well, all states are sent together anyway, this might make sense for the requests (atomicity). But they might be ignored, does it still make sense to be atomic?

unstable/wlr-foreign-toplevel-list.xml
emersion: After the server sends this event, it might receive more client requests (that the client has sent before receiving the event). The server can't guess whether the client has sent those before receiving the event or after (so making this a protocol error is racy).

ammen99: I see what you mean.

unstable/wlr-foreign-toplevel-list.xml
ammen99: this is still unresolved, not sure how to make github display it properly

ammen99:
later as in a few hours

SirCmpwn:
:+1:

ammen99:
I have added a simple done event, because that's how it is called in xdg-output and seems closest to our purpose.

I have several questions:

1. Do we list all events that are included in the done event (output_enter/leave, state ..)?
2. destroyed event shouldn't be included in any case, right? I don't think it makes sense for any kind "atomicity" when the destroyed event is concerned
3. Do we require that done was sent before the destroyed?

ammen99:
bump? anything else I should consider?

SirCmpwn:
You should probably start writing an example client and server implementation to prove this out.

ammen99:
Sorry for the long delay, but I'had to focus on other things. I updated the protocol and the implementation to include the proposed seat argument, thanks @nyorain

ammen99:
bump?

unstable/wlr-foreign-toplevel-unstable-v1.xml
```diff
@@ -0,0 +1,229 @@
...
+  </copyright>
+
+  <interface name="zwlr_foreign_toplevel_manager_v1" version="1">
+    <description summary="list and control opened apps">
+      The purpose of this protocol is to enable the creation of taskbars
+      and docks by providing them with a list of opened applications and
+      letting them request certain actions on them, like maximizing, etc.
+
+      After a client binds the zwlr_foreign_toplevel_manager_v1, each opened
+      toplevel window will be sent via the toplevel_created signal.
```
emersion: toplevel_created → toplevel

signal → event

unstable/wlr-foreign-toplevel-unstable-v1.xml
```diff
@@ -0,0 +1,229 @@
...
+      <description summary="stop sending events">
+        Indicates the client no longer wishes to receive events for new toplevels.
+        However the compositor may emit further toplevel_created events, until
+        the finished event is emitted.
+
+        The client must not send any more requests after this one.
+      </description>
+    </request>
+
+    <event name="finished">
```
emersion: This event should be marked as destructor.

unstable/wlr-foreign-toplevel-unstable-v1.xml
```diff
@@ -0,0 +1,229 @@
...
+      </description>
+
+      <arg name="surface" type="object" interface="wl_surface"/>
+      <arg name="x" type="int"/>
+      <arg name="y" type="int"/>
+      <arg name="width" type="int"/>
+      <arg name="height" type="int"/>
+    </request>
+
+    <event name="destroyed">
```
emersion: destroyed → closed?

emersion:
Overall, I'm not quite sure I like this design. I expect this protocol to be used for things like screenshooting and screencasting too, and I don't like how things like `set_minimized` are exposed here.

I'd prefer having two protocols. One protocol which just lists toplevels, and another one which extends foreign toplevels with requests like `set_minimized`.

SirCmpwn:
This protocol is designed to serve taskbars more so than screenshots. Even so, I like your proposed split.

unstable/wlr-foreign-toplevel-unstable-v1.xml
emersion: The fact that you need to call `destroy` on a `destroyed` object is confusing.

ammen99: maybe "done" then?

emersion: Nah, `done` is used as the compositor-side version of `commit`. See for instance `wl_output`.

ammen99:
As discussed in IRC, this protocol needs to be renamed. Does the name `wlr-toplevel-management` (or `wlr-foreign-toplevel-management`, but I find that a bit too long) make sense? Then the `wlr_toplevel` interface to `wlr-toplevel-handle`?

If later a generic toplevel listing protocol is implemented, it could be called `wlr-toplevel-list` with `wlr-toplevel` representing the individual toplevel, without naming clashes. Plus, a 'handle' is AFAIK generally something you can use to control something, making it a good choice for this protocol.

ddevault:
I like `wlr-foreign-toplevel-management`.

unstable/wlr-foreign-toplevel-management-unstable-v1.xml
```diff
@@ -0,0 +1,230 @@
...
+      <arg name="toplevel" type="new_id" interface="zwlr_foreign_toplevel_handle_v1"/>
+    </event>
+
+    <request name="stop">
+      <description summary="stop sending events">
+        Indicates the client no longer wishes to receive events for new toplevels.
+        However the compositor may emit further toplevel_created events, until
+        the finished event is emitted.
+
+        The client must not send any more requests after this one.
```
emersion: We should add an error code for this.

unstable/wlr-foreign-toplevel-management-unstable-v1.xml
```diff
@@ -0,0 +1,230 @@
...
+        the place where the app using this protocol represents the given toplevel.
+        It can be used by the compositor as a hint for some operations, e.g
+        minimizing. The client is however not required to set this, in which
+        case the compositor is free to decide some default value.
+
+        If the client specifies more than one rectangle, only the last one is
+        considered.
+
+        The dimensions are given in surface-local coordinates.
+        Setting width=height=0 removes the already-set rectangle.
```
emersion: Add an error for invalid values.

unstable/wlr-foreign-toplevel-management-unstable-v1.xml
ammen99: it should be noted that we (in wlroots) destroy the resource server-side immediately when receiving this. How is the error handled in this case, won't libwayland drop the request or error the client?

emersion: Hmm, you're right. It's probably fine then.

emersion:
Thanks!

PR #28, opened 2018-10-12
Add Makefile and pkg-config entry

### LINKED ISSUE #8: Make wlr-protocols into an installable package
Similar to how wayland-protocols is an installable package, we should investigate whether or not it would be better to have this as an installable package as well.

This way people could bring it in as a dependency into their projects. The alternative is to have people copy it over which makes updates hard and subjects us to protocol drift.

Required:

* Gather feedback from wlr-protocols users to determine whether or not this is a good idea.
* Add a build system (meson or autotools) that simply installs the protocols probably in some directory in `/usr/share`.

SirCmpwn:
Closes #8

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,38 @@
+PREFIX=/usr
+DATADIR=$${datarootdir}
+DATAROOTDIR=$${prefix}/share
+
+unstable_protocols = \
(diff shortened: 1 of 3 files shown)


emersion:
Thanks!

PR #31, opened 2018-11-18
Add wlr-keyboard-modifiers-unstable protocol

harishkrupo:
Signed-off-by: Harish Krupo <harish.krupo.kps@intel.com>

### DIFF
--- a/unstable/wlr-keyboard-modifiers-unstable-v1.xml
+++ b/unstable/wlr-keyboard-modifiers-unstable-v1.xml
@@ -1,0 +1,70 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<protocol name="wlr_keyboard_modifiers_unstable_v1">
+  <copyright>
+    Copyright © 2018 Harish Krupo
+
+    Permission to use, copy, modify, distribute, and sell this
+    software and its documentation for any purpose is hereby granted
+    without fee, provided that the above copyright notice appear in
+    all copies and that both that copyright notice and this permission
+    notice appear in supporting documentation, and that the name of
+    the copyright holders not be used in advertising or publicity
+    pertaining to distribution of the software without specific,
+    written prior permission.  The copyright holders make no
+    representations about the suitability of this software for any
+    purpose.  It is provided "as is" without express or implied
+    warranty.
+
+    THE COPYRIGHT HOLDERS DISCLAIM ALL WARRANTIES WITH REGARD TO THIS
+    SOFTWARE, INCLUDING ALL IMPLIED WARRANTIES OF MERCHANTABILITY AND
+    FITNESS, IN NO EVENT SHALL THE COPYRIGHT HOLDERS BE LIABLE FOR ANY
+    SPECIAL, INDIRECT OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN
+    AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION,
+    ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF
+    THIS SOFTWARE.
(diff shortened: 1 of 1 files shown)


harishkrupo:
> I think I'm still -1 on this protocol tbh.

Do you have any suggestions to improve it? 
Any other way to get modifiers then?

bemeurer:
I'd just like to add my support for this as a user. I constantly switch between three languages that require distinct layouts, and it's super helpful to have my status bar display the current one. I get that this protocol might seem too specific, but I feel like it addresses an important point. Thanks @harishkrupo for putting in the work for this, and @emersion and @ddevault for giving feedback and guidance!

AdrianVovk:
As a DE dev, I would also like to vouch for the usefulness of this protocol.

gdamjan:
the protocol should probably be extended to get all the configured layouts (so that you can have a pop-up menu with all the layouts), and also to switch to another layout.

ddevault:
>I'd just like to add my support for this as a user. I constantly switch between three languages that require distinct layouts, and it's super helpful to have my status bar display the current one. I get that this protocol might seem too specific, but I feel like it addresses an important point. Thanks @harishkrupo for putting in the work for this, and @emersion and @ddevault for giving feedback and guidance!

￼￼afaict this protocol does not fulfill this use-case. But you can fulfill this use-case anyway by just listening for keymaps on `wl_seat`, right?

harishkrupo:
> ￼￼afaict this protocol does not fulfill this use-case. But you can fulfill this use-case anyway by just listening for keymaps on `wl_seat`, right?

@ddevault, Does wl_seat advertise keymaps? I don't see any event for that here: https://gitlab.freedesktop.org/wayland/wayland/blob/master/protocol/wayland.xml#L1663

gdamjan:
> Enumerating all layouts is possible client-side using the libxkbcommon API. It gives names like English (US) though, not things like en.

https://github.com/xkbcommon/libxkbcommon/issues/22
people suggested parsing evdev.xml (part of xkeyboard-config)

emersion:
>people suggested parsing evdev.xml (part of xkeyboard-config)

All right. In any case, it doesn't belong to this protocol, and should be done client-side.

ammen99:
Does the xkb state sent by the compositor also contain the list of enabled layouts?

emersion:
>Does the xkb state sent by the compositor also contain the list of enabled layouts?

Yes.

agx:
Extending this a bit: what if a client (bar or shell) wants to listen to e.g. KEY_VOLUMEDOWN, KEY_VOLUMEUP or KEY_POWER. If we want to allow clients to handle this we might need a protocol to send certain input events to special clients (maybe using a whitelist in the compositor so clients can request certain input events but only get access to them if whitelisted).

emersion:
>Extending this a bit: what if a client (bar or shell) wants to listen to e.g. KEY_VOLUMEDOWN, KEY_VOLUMEUP or KEY_POWER. If we want to allow clients to handle this we might need a protocol to send certain input events to special clients (maybe using a whitelist in the compositor so clients can request certain input events but only get access to them if whitelisted).

Related: https://patchwork.freedesktop.org/patch/172113/

unstable/wlr-keyboard-modifiers-unstable-v1.xml
```diff
@@ -0,0 +1,70 @@
...
+  </copyright>
+
+  <interface name="zwlr_keyboard_modifiers_v1" version="1">
+    <description summary="lets client recieve the keyboard modifiers">
+
+    </description>
+
+    <request name="get_wlr_keyboard">
+      <description summary="Create a wrapper over wl_keyboard">
+        Retruns a wrapper for the wl_keyboard passed as an argument. The
```
craigbarnes, line 35:
```suggestion
        Returns a wrapper for the wl_keyboard passed as an argument. The
```

emersion:

https://gitlab.freedesktop.org/wlroots/wlr-protocols/-/merge_requests/31

PR #34, opened 2019-01-29
Support primary selection in data-control

YaLTeR:
These changes add support for the primary selection to the `data-control` protocol, alongside the regular selection. With these changes, clipboard managers (or tools like [wl-clipboard](https://github.com/bugaevc/wl-clipboard)) can work with both regular and primary clipboard contents without refraining to hacks like spawning invisible surfaces and stealing focus. See https://github.com/bugaevc/wl-clipboard/issues/24.

With these changes, invalidation of data offers is no longer "automatic" upon receiving a new data offer or selection, and instead should be controlled by the compositor using the new `wlr_data_control_offer_v1.finished` event. This is to support
- two different data offers (one for the regular clipboard and one for the primary clipboard);
- one data offer shared between the regular and the primary clipboard (think tools like [xcutsel](https://linux.die.net/man/1/xcutsel) for synchronizing regular and primary clipboards: it's possible to simply receive the regular clipboard data offer and set it as the primary selection too without having to copy it).

cc @bugaevc @emersion

### DIFF
--- a/unstable/wlr-data-control-unstable-v1.xml
+++ b/unstable/wlr-data-control-unstable-v1.xml
@@ -103,9 +103,6 @@
         event, the new data_offer object will send out
         wlr_data_control_offer.offer events to describe the MIME types it
         offers.
-
-        This event replaces the previous data offer, which should be destroyed
-        by the client.
       </description>
       <arg name="id" type="new_id" interface="zwlr_data_control_offer_v1"/>
     </event>
--- a/unstable/wlr-data-control-unstable-v1.xml
+++ b/unstable/wlr-data-control-unstable-v1.xml
@@ -111,7 +111,7 @@
     </event>
 
     <event name="selection">
-      <description summary="introduce a new wlr_data_control_offer">
+      <description summary="advertise new selection">
         The selection event is sent out to notify the client of a new
         wlr_data_control_offer for the selection for this device. The
         wlr_data_control_device.data_offer and the wlr_data_control_offer.offer
--- a/unstable/wlr-data-control-unstable-v1.xml
+++ b/unstable/wlr-data-control-unstable-v1.xml
@@ -2,6 +2,7 @@
 <protocol name="wlr_data_control_unstable_v1">
   <copyright>
     Copyright © 2018 Simon Ser
+    Copyright © 2019 Ivan Molodetskikh
 
     Permission to use, copy, modify, distribute, and sell this
     software and its documentation for any purpose is hereby granted
@@ -79,8 +80,10 @@
 
     <request name="set_selection">
       <description summary="copy data to the selection">
-        All objects created by the device will still remain valid, until their
-        appropriate destroy request has been called.
+        This request asks the compositor to set the selection to the data from
+        the source on behalf of the client.
+
+        To unset the selection, set the source to NULL.
       </description>
       <arg name="source" type="object" interface="zwlr_data_control_source_v1"
         allow-null="true"/>
--- a/unstable/wlr-data-control-unstable-v1.xml
+++ b/unstable/wlr-data-control-unstable-v1.xml
@@ -83,6 +83,9 @@
         This request asks the compositor to set the selection to the data from
         the source on behalf of the client.
 
+        The given source may not be used in any further set_selection or
+        set_primary_selection requests.
+
         To unset the selection, set the source to NULL.
       </description>
       <arg name="source" type="object" interface="zwlr_data_control_source_v1"
@@ -161,6 +164,9 @@
       <description summary="copy data to the primary selection">
         This request asks the compositor to set the primary selection to the
         data from the source on behalf of the client.
+
+        The given source may not be used in any further set_selection or
+        set_primary_selection requests.
 
         To unset the primary selection, set the source to NULL.
 
--- a/unstable/wlr-data-control-unstable-v1.xml
+++ b/unstable/wlr-data-control-unstable-v1.xml
@@ -84,7 +84,8 @@
         the source on behalf of the client.
 
         The given source may not be used in any further set_selection or
-        set_primary_selection requests.
+        set_primary_selection requests. Attempting to use a previously used
+        source is a protocol error.
 
         To unset the selection, set the source to NULL.
       </description>
@@ -166,7 +167,8 @@
         data from the source on behalf of the client.
 
         The given source may not be used in any further set_selection or
-        set_primary_selection requests.
+        set_primary_selection requests. Attempting to use a previously used
+        source is a protocol error.
 
         To unset the primary selection, set the source to NULL.
 
@@ -176,6 +178,11 @@
       <arg name="source" type="object" interface="zwlr_data_control_source_v1"
         allow-null="true"/>
     </request>
+
+    <enum name="error" since="2">
+      <entry name="used_source" value="1"
+        summary="source given to set_selection or set_primary_selection was already used before"/>
+    </enum>
   </interface>
 
   <interface name="zwlr_data_control_source_v1" version="1">
--- a/unstable/wlr-data-control-unstable-v1.xml
+++ b/unstable/wlr-data-control-unstable-v1.xml
@@ -41,7 +41,7 @@
     interface version number is reset.
   </description>
 
-  <interface name="zwlr_data_control_manager_v1" version="1">
+  <interface name="zwlr_data_control_manager_v1" version="2">
     <description summary="manager to control data devices">
       This interface is a manager that allows creating per-seat data device
       controls.
@@ -71,7 +71,7 @@
     </request>
   </interface>
 
-  <interface name="zwlr_data_control_device_v1" version="1">
+  <interface name="zwlr_data_control_device_v1" version="2">
     <description summary="manage a data device for a seat">
       This interface allows a client to manage a seat's selection.
 
@@ -98,11 +98,13 @@
     <event name="data_offer">
       <description summary="introduce a new wlr_data_control_offer">
         The data_offer event introduces a new wlr_data_control_offer object,
-        which will subsequently be used in the wlr_data_control_device.selection
-        event. Immediately following the wlr_data_control_device.data_offer
-        event, the new data_offer object will send out
-        wlr_data_control_offer.offer events to describe the MIME types it
-        offers.
+        which will subsequently be used in either the
+        wlr_data_control_device.selection event (for the regular clipboard
+        selections) or the wlr_data_control_device.primary_selection event (for
+        the primary clipboard selections). Immediately following the
+        wlr_data_control_device.data_offer event, the new data_offer object
+        will send out wlr_data_control_offer.offer events to describe the MIME
+        types it offers.
       </description>
       <arg name="id" type="new_id" interface="zwlr_data_control_offer_v1"/>
     </event>
@@ -129,6 +131,38 @@
         the client.
       </description>
     </event>
+
+    <!-- Version 2 additions -->
+
+    <event name="primary_selection" since="2">
+      <description summary="advertise new primary selection">
+        The primary_selection event is sent out to notify the client of a new
+        wlr_data_control_offer for the primary selection for this device. The
+        wlr_data_control_device.data_offer and the wlr_data_control_offer.offer
+        events are sent out immediately before this event to introduce the data
+        offer object. The primary_selection event is sent to a client when a
+        new primary selection is set. The wlr_data_control_offer is valid until
+        a new wlr_data_control_offer or NULL is received. The client must
+        destroy the previous primary selection wlr_data_control_offer, if any,
+        upon receiving this event.
+      </description>
+      <arg name="id" type="object" interface="zwlr_data_control_offer_v1"
+        allow-null="true"/>
+    </event>
+
+    <request name="set_primary_selection" since="2">
+      <description summary="copy data to the primary selection">
+        This request asks the compositor to set the primary selection to the
+        data from the source on behalf of the client.
+
+        To unset the primary selection, set the source to NULL.
+
+        The compositor will ignore this request if it does not support primary
+        selection.
+      </description>
+      <arg name="source" type="object" interface="zwlr_data_control_source_v1"
+        allow-null="true"/>
+    </request>
   </interface>
 
   <interface name="zwlr_data_control_source_v1" version="1">
--- a/unstable/wlr-data-control-unstable-v1.xml
+++ b/unstable/wlr-data-control-unstable-v1.xml
@@ -120,6 +120,9 @@
         wlr_data_control_offer or NULL is received. The client must destroy the
         previous selection wlr_data_control_offer, if any, upon receiving this
         event.
+
+        The first selection event is sent upon binding the
+        wlr_data_control_device object.
       </description>
       <arg name="id" type="object" interface="zwlr_data_control_offer_v1"
         allow-null="true"/>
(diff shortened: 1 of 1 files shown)


unstable/wlr-data-control-unstable-v1.xml
```diff
@@ -112,12 +116,17 @@
...
+        set.
+
+        For objects of version 1, the wlr_data_control_offer is valid until a new
         wlr_data_control_offer or NULL is received. The client must destroy the
-        previous selection wlr_data_control_offer, if any, upon receiving this
+        previous primary selection wlr_data_control_offer, if any, upon receiving this
         event.
+
+        Starting from version 2, this event does not invalidate the previously
+        received wlr_data_control_offer objects.
```
emersion: Can't the client rely on the `selection` and `primary_selection` events instead? This would be more consistent with `wl_data_device` (which has selection offers vs. drag-and-drop offers).

unstable/wlr-data-control-unstable-v1.xml
```diff
@@ -129,6 +138,30 @@
...
+        primary selection is set.
+      </description>
+      <arg name="id" type="object" interface="zwlr_data_control_offer_v1"
+        allow-null="true"/>
+    </event>
+
+    <request name="set_primary_selection" since="2">
+      <description summary="copy data to the primary selection">
+        All objects created by the device will still remain valid, until their
+        appropriate destroy request has been called.
```
emersion: It would be nice to have a note about primary-selection support. Not all compositors support primary-selection, so we should make it clear that this request may be ignored by the compositor.

In the general case, even if primary-selection is supported, the compositor is free to ignore this request.

Additionally, it seems that the request description is a little bit off.

unstable/wlr-data-control-unstable-v1.xml
YaLTeR: What if both the primary and the regular clipboard contain the same data offer?

emersion: I think we should re-use the core protocol wording:

>The data_offer event introduces a new wl_data_offer object, which will subsequently be used in either the data_device.enter event (for drag-and-drop) or the data_device.selection event (for selections).

YaLTeR: So if a client creates a data offer and then sends it to both `selection` and `primary_selection`, other clients will receive two separate data offers? That would allow keeping the previous invalidate-previous-upon-selection-change semantics.

YaLTeR:
There's also [this suggestion](https://github.com/swaywm/wlr-protocols/pull/25#issuecomment-444482218) on adding a capabilities-type event for advertising whether the compositor supports the primary selection. Should I add that too?

unstable/wlr-data-control-unstable-v1.xml
emersion: Clients don't create offers. Clients create sources, then calls `set_selection` or `set_primary_selection`, and the compositor will send offers to focused clients via the `selection` and `primary_selection` events.

emersion: >other clients will receive two separate data offers?

Yes, that is the case.

emersion:
> There's also this suggestion on adding a capabilities-type event for advertising whether the compositor supports the primary selection. Should I add that too?

I think support for primary selection can be inferred from the supported protocols (if the primary-selection global is exposed, then the compositor supports it). I'd prefer not to add it if we don't have a clear use-case for it.

unstable/wlr-data-control-unstable-v1.xml
YaLTeR: > Clients don't create offers. Clients create sources

Ah, right, forgot about that.

unstable/wlr-data-control-unstable-v1.xml
```diff
@@ -95,20 +98,25 @@
...
-        offers.
+        which will subsequently be used in either the
+        wlr_data_control_device.selection event (for the regular clipboard
+        selections) or the wlr_data_control_device.primary_selection event (for
+        the primary clipboard selections). Immediately following the
+        wlr_data_control_device.data_offer event, the new data_offer object
+        will send out wlr_data_control_offer.offer events to describe the MIME
+        types it offers.
+
+        For objects of version 1, this event replaces the previous data offer,
```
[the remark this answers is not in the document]
YaLTeR: I was confused by this wording too when reading through `data-control` for the first time. Do you suggest removing it altogether (so even for v1 only `selection` can invalidate the data offers and not new data offers by themselves)?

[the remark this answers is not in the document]
emersion: Yes, I think it makes a lot more sense.

bugaevc:
> I think support for primary selection can be inferred from the supported protocols (if the primary-selection global is exposed, then the compositor supports it). I'd prefer not to add it if we don't have a clear use-case for it.

Do you mean the client should check for a global named `zwp_primary_selection_device_manager_v1`? What if the compositor supports primary selection via gtk-primary-selection and wlr-data-control but does not support wp-primary-selection? What if it supports a different (future) version of wp-primary-selection that uses a different global name (`zwp_primary_selection_device_manager_v2`) that a client doesn't know about, are clients expected to guess which global names indicate primary selection support (as in `strstr("primary")`)?

Unless I misunderstand what you mean, I'd strongly prefer wlr-data-control to provide an explicit way for the client to know what's supported.

emersion:
>Unless I misunderstand what you mean, I'd strongly prefer wlr-data-control to provide an explicit way for the client to know what's supported.

What's your use-case for this?

bugaevc:
My exact use case would be this:

wl-clipboard can be built without wp-primary-selection support (which is what happens if `dependency('wayland-protocols')` is not found or its version is not `>= 1.17`), in which case it knows nothing about wp-primary-selection. Yet it needs to know whether getting/setting primary selection via wlr-data-control is enough or it needs to pop up a surface and hope to get/set primary selection via gtk-primary-selection.

I guess I _could_ still check for the global name even if the protocol is not supported... and if gtk-primary-selection is supported that probably means wlr-data-control supports primary selection too (currently in wl-clipboard, wlr-data-control and gtk-primary-selection are both either supported or not supported — though that could change...), but why add another hack when we could make the protocol clearly state what it supports.

YaLTeR:
I assume something like `wl-copy` would benefit from knowing definitely whether `data-control` supports primary selection so it can return an error to the user when trying to copy into the unsupported primary selection instead of just blindly sending data to be ignored by the compositor (leaving the user wondering why `wl-paste` reports nothing in the primary selection when `wl-copy` clearly copied the data without errors).

emersion:
Makes sense. I guess we could add a `primary_selection` event to `zwlr_data_control_manager_v1`, immediately sent after the client binds to the global, to advertise for primary-selection support.

YaLTeR:
Another possible improvement: it seems to me that it is currently unspecified when the compositor will send the first `selection` event. Perhaps it should be specified that the first `selection` event will be sent immediately after binding the `wlr_data_control_device` (either with NULL or a data offer), and that the first `primary_selection` event will be either not sent at all (if the primary selection support is not advertised) or will also be sent immediately after binding (or maybe it can be sent either way, with guaranteed NULL if not supported)?

unstable/wlr-data-control-unstable-v1.xml
```diff
@@ -68,9 +69,18 @@
         appropriate destroy request has been called.
       </description>
     </request>
+
+    <!-- Version 2 additions -->
+
+    <event name="primary_selection" since="2">
+      <description summary="advertise primary selection support">
+        This event is sent when binding to the wlr_data_control_manager global
```
emersion: s/is/can be/

ddevault:
Can you rebase this into a more coherent history?

YaLTeR:
Added clarifications on the first selection events and rebased into a coherent history.

emersion:
This looks good to me. I think we can move on to the client+server implementations.

bugaevc:
> This looks good to me. I think we can move on to the client+server implementations.

Seconded. I'll try to implement it in wl-clipboard tomorrow.

bugaevc:
https://github.com/bugaevc/wl-clipboard/tree/data-control-primary

It may or may not work (I have nothing to test it against...), but at least it seems to handle the cases when the compositor does not support wlr-data-control v2 correctly.

YaLTeR:
Update as per https://github.com/swaywm/wlroots/pull/1520#issuecomment-459503055.

YaLTeR:
Hm, using a data source multiple times should probably result in a protocol error. Should I add an error enum to `data_control_device` then? How does that work with versioning?

emersion:
>Hm, using a data source multiple times should probably result in a protocol error. Should I add an error enum to data_control_device then? How does that work with versioning?

+1

The error enum should be added to the data device though. It can have a `since` attribute.

YaLTeR:
So for protocol v1 the compositor silently ignores extra `set_selection`s and for protocol v2 it throws an error?

YaLTeR:
Looks like there's no `SINCE` define generated for that `since="2"` on the enum.

emersion:
>So for protocol v1 the compositor silently ignores extra set_selections and for protocol v2 it throws an error?

Ignoring looks like a bad idea.

>Looks like there's no SINCE define generated for that since="2" on the enum.

No big deal, sending an unknown error for v1 is not so bad.

unstable/wlr-data-control-unstable-v1.xml
```diff
@@ -68,9 +69,18 @@
...
     </request>
+
+    <!-- Version 2 additions -->
+
+    <event name="primary_selection" since="2">
+      <description summary="advertise primary selection support">
+        This event can be sent when binding to the wlr_data_control_manager
+        global to advertise that it supports the primary selection.
+      </description>
+    </event>
```
ddevault: This was probably discussed and I'm probably dumb for bringing it up

But isn't the existence of the primary selection global enough of an advertisement?

bugaevc: It was discussed somewhat (see above). In brief, the problems with this is that there's no _the_ primary selection global; there are two protocols that implement primary selection as of now and there may be more in the future. The compositor may (only) support e.g. `zwp_primvary_selection_v5` whereas a tool that intends to use this only knows about `gtk_primary_selection`. The tools would have to resort to hacks like `strstr("primary")` which is relatively ugly compared to wlr-data-control stating what it supports explicitly.

ddevault: I consider gtk_primary_selection deprecated and all clients/compositors should be moving towards zwp. Future improvements to primary selection should land in zwp as well. We don't need to consider the needs of gtk_primary_selection here at all, I think.

bugaevc: It's not about `gtk_primary_selection` itself, it's about what happens if the client and the compositor don't have a match in the (version of) the primary selection protocol they support. They could still both support wlr-data-control and do primary selection stuff through that, but there'd be no (sane) way for them to know.

And this is not a theoretical possibility; wl-clipboard doesn't support or know about `wp_primary_selection` if it's built with `wayland-protocols` v1.16 or earlier (which is what current Linux distros ship); if you remove `gtk_primary_selection` from Sway we're going to get into this very situation.

ddevault: I think the solution in that case is to send a patch updating whichever end is out of date. @emersion thoughts?

emersion: More context here: https://github.com/swaywm/wlr-protocols/pull/34#issuecomment-458548155

This event makes error reporting friendlier, but at the same time the compositor could've decided not to set the selection for whatever reason. I'm fine with both tbh.

ddevault: My vote is in favor of removing this.

bugaevc: Supposing this ends up being removed, what exactly would you recommend a client wishing to know whether this protocol is suitable for managing primary selection to do?

For example, you could say that one can manage primary selection using this protocol _if and only if_ there is a global named `wp_primary_selection_unstable_v1` (or something). In which case it'd be better if this fact was mentioned in the protocol.

YaLTeR: I don't see another reliable way of figuring out whether the compositor supports primary selection. Besides, why would we _want_ to rely on hacks when this event (which falls in scope of the protocol) solves the issue?

ddevault: Look for the global. This isn't a hack.

bugaevc: _Which_ global is "the" global? `wp_primary_selection_unstable_v1`? `gtk_primary_selection` _or_ `wp_primary_selection_unstable_v1`? _Any_ global whose name includes "primary"?

I still thinks it's a bad idea, but if you decide to do it this way, you'll have to document it in the protocol, perhaps like this:

```
This request can only be made if a global with name wp_primary_selection_unstable_v1 is available. Otherwise, a protocol error will be raised.
```

or

```
This request will only have any effect if a global with name wp_primary_selection_unstable_v1 is available. Otherwise, it will be silently ignored.
```

ddevault: We already went over this. wp-primary-selection, whichever is the latest version.

ddevault: And we can just silently ignore it if the client makes the request, we needn't document that.

YaLTeR: Just to be clear, let me restate the points.

Utilities like `wl-copy` and `wl-paste` and other clipboard managing tools want to know whether the primary selection is supported to show an error message in case it isn't, instead of silently setting or retrieving an empty selection (which would be confusing to the user).

The `primary_selection` event in `data-control` is one way of solving this issue.

Your suggestion is that the presence of a common primary selection protocol global should be used instead. At the current time that would be `zwp_primary_selection_device_manager_v1`.

However, we argue that it isn't optimal, because when `wp-primary-selection` gets a breaking change or stabilizes, the name of the global will change. This would break any clipboard manager's primary selection support until the new name is added into its checks, a new release is drafted, is picked up by distributions, etc. Essentially it means that clipboard utilities relying on `data-control` will break and need a new release due to an update of a protocol that they _don't even use_.

ddevault: Yes, because if the unstable version changes, it's a different protocol entirely. Once it becomes stable the global name will not change. This is by design.

YaLTeR: The issue is that clipboard managers would only care about the _presence_ of the `wp-primary-selection` main global as an indicator for primary selection support (_not_ for operation, clipboard managers use `data-control` for all operation). Breaking changes of `wp-primary-selection` in no way affect the clipboard managers, yet under your proposal they would still require an update and a new release if `wp-primary-selection` has a breaking change or stabilizes.

ddevault: I've said my piece. Maybe @emersion can make the call.

emersion: Hmm. In the end I'd prefer removing this. The long-term solution will be to check for `wp-primary-selection`'s global, so this event won't be needed. In the meantime you can add a check for `zwp-primary-selection-v*`. Also, if there is any major issue with this approach it's always easier to add a new event than to remove it.

YaLTeR:
Removed `wlr_data_control_manager.primary_selection`. Also I just realized, since we state that `wlr_data_control_device.primary_selection` is sent out immediately on binding, if the primary selection is supported, that can probably be used to tell if the compositor supports primary selection (got one => supports, didn't get one on the first roundtrip => doesn't support).

emersion:
Seems like removing [`</interface>`](https://github.com/swaywm/wlr-protocols/pull/34/files#diff-e943889b2de92a9ed11afe12695a9210L71) fails the build: https://builds.sr.ht/~emersion/job/29777#task-protocols

YaLTeR:
> Seems like removing [`</interface>`](https://github.com/swaywm/wlr-protocols/pull/34/files#diff-e943889b2de92a9ed11afe12695a9210L71) fails the build: https://builds.sr.ht/~emersion/job/29777#task-protocols

Whoops. Rebased to simply remove the commit that added the event.

ddevault:
Thanks!

PR #35, opened 2019-02-25
WIP: Added wlr-workspace-unstable-v1.xml

chrisjbillington:
This is a draft for addressing issue #10.

I likely don't have the C chops to contribute much to wlroots' code (though I might try), but maybe I can talk about protocols.

To accommodate both compositors that have workspaces spanning all outputs, as well as those that have each output with its own independent set of workspaces, this protocol has the concept of 'workspace groups'. A workspace group is a set of outputs that share workspaces. Each output may belong to at most one group. Each workspace group has one active workspace at any one time.  This allows arbitrary grouping of outputs, including the two common cases of all outputs sharing a group and of each output being its own group.

There is more to think about here.

Is the protocol sufficient for required atomicity? If an output moves to a different workspace group, then the 'each output may belong to at most one group' requirement will be violated from the perspective of the client until both an `output` and an `output_remove` event are processed on the two `zwlr_workspace_group_handle_v1` objects. But the compositor can refrain from sending `done` events until both of these events are sent, so we could say that if the client gets an `output` or `output_remove` event, it must wait for the `done` event before it has a consistent state for what groups all the outputs belong to.

Similarly with `zwlr_workspace_handle_v1`, if a workspace changes what its number is, the client must wait for the `done` event in order to see that the 'workspaces are sequentially numbered from zero' requirement is satisfied.

I am also a bit confused about destroy events and requests. If you request to destroy an object, you're just asking to clean up resources - you're not asking the compositor to remove a workspace or whatnot. However, if the compositor *does* remove a workspace (not just move it to a different workspace group), does it send a `destroyed` event?

I suspect the general ideas in this protocol are pretty unobjectionable, but potentially more contentious things that I haven't included are:

Workspaces are simply numbered here. Should it be left that way, or should they be defined on a 2D grid? If we do not allow for information about a 2D grid, then if the compositor is laying out the workspaces this way, panels will not be able to reflect that in how they display workspaces, and they will not be able to implement 'move to workspace left' and 'move to workspace up' functionality, etc.

Should workspaces and/or workspace groups have names?

Here are some events/requests I have not included

zwlr_workspace_manager_v1:
* request: create workspace group

zwlr_workspace_group_v1:
* request: add an output to the group
* request: remove an output from the group
* request: add a workspace to the group
* request: remove a workspace from the group
* request: set a name as a string
* event: name as a string

zwl_workspace_v1:
* event: name as a string
* request: move (change the number of the workspace - processed as an insert to keep numbers sequential)
* request: set name as a string

There is also some question of whether only one workspace can be active at a time. In compositor workspace overview interfaces, you can often see multiple workspaces. However, that's not inconsistent with only one of them being 'active', active doesn't have to mean visible, so I'm not sure if that's a problem.

### DIFF
--- a/unstable/wlr-workspace-unstable-v1.xml
+++ b/unstable/wlr-workspace-unstable-v1.xml
@@ -30,15 +30,17 @@
       Workspaces, also called virtual desktops, are groups of surfaces. A compositor
       with a concept of workspaces may only show some such groups of surfaces (those of
       'active' workspaces) at a time. Workspaces may contain surfaces from one or more
-      outputs, such that activating the workspace will request that the compositor
-      display the workspace's surfaces on those outputs, whereas the compositor may hide
-      or otherwise de-emphasise surfaces that are associated only with inactive
+      outputs, and 'activating' a workspace is a request that the compositor display
+      those surfaces on those outputs as normal, whereas the compositor may hide or
+      otherwise de-emphasise surfaces that are associated only with 'inactive'
       workspaces. By associating a set of outputs with each workspace, each output may
       have its own unique set of workspaces, or all outputs (or any other arbitrary
-      grouping) may share workspaces. Workspaces may be conceptually arranged in a grid
-      of arbitrary dimension. The purpose of this protocol is to enable the creation of
-      taskbars and docks by providing them with a list of workspaces and their
-      properties, and allowing them to activate and deactivate workspaces.
+      grouping) may share workspaces. Compositors may optionally conceptually arrange
+      workspaces with a common set of outputs in an N-dimensional grid.
+
+      The purpose of this protocol is to enable the creation of taskbars and docks by
+      providing them with a list of workspaces and their properties, and allowing them
+      to activate and deactivate workspaces.
 
       After a client binds the zwlr_workspace_manager_v1, each workspace will be sent
       via the workspace event.
@@ -87,14 +89,14 @@
       activate or deactivate the workspace.
     </description>
 
-    <event name="output">
+    <event name="enter">
       <description summary="output added to workspace">
         This event is emitted whenever an output is added to the workspace.
       </description>
       <arg name="output" type="object" interface="wl_output"/>
     </event>
 
-    <event name="output_remove">
+    <event name="leave">
       <description summary="output removed from workspace">
         This event is emitted whenever an output is removed from the workspace.
       </description>
@@ -106,19 +108,27 @@
         This event is emitted immediately after the zwlr_workspace_handle_v1 is created
         and whenever the name of the workspace changes.
       </description>
-      <arg name="number" type="int"/>
+      <arg name="name" type="string"/>
     </event>
 
     <event name="coordinates">
       <description summary="workspace coordinates changed">
-        If the compositor assigns coordinates to workspaces, this event is emitted
-        immediately after the zwlr_workspace_handle_v1 is created and whenever the
-        coordinates of the workspace change. Coordinates have an arbitrary number of
-        dimensions with an unsigned integer position for each dimension. Among a set of
-        workspaces that have identical sets of outputs, workspaces must have unique
-        coordinates with the same number of dimensions, such that the workspaces form a
-        grid with at most one workspace in each position. By convention, 2D coordinates
-        represent [y_pos, x_pos].
+        This event is used to organize workspaces spatially into an N-dimensional grid,
+        and if supported, is emitted immediately after the zwlr_workspace_handle_v1 is
+        created and whenever the coordinates of the workspace change. Compositors may
+        not send this event if they do not conceptually arrange workspaces in this way.
+
+        Coordinates have an arbitrary number of dimensions with an uint32 position along
+        each dimension. By convention, the first dimension is X, the second Y, the third
+        Z, and so on. The compositor may chose to utilize these events for a more novel
+        workspace layout convention, however. No guarantee is made about the grid being
+        filled or bounded; there may be a workspace at coordinate 1 and another at
+        coordinate 1000 and none in between.
+
+        Among a set of workspaces that have identical sets of outputs, workspaces must
+        have unique coordinates of the same dimensionality. Coordinates otherwise need
+        not be unique: in this way independent grids can exist for each set of outputs
+        that share workspaces.
       </description>
       <arg name="coordinates" type="array"/>
     </event>
--- a/unstable/wlr-workspace-unstable-v1.xml
+++ b/unstable/wlr-workspace-unstable-v1.xml
@@ -117,13 +117,16 @@
         and if supported, is emitted immediately after the zwlr_workspace_handle_v1 is
         created and whenever the coordinates of the workspace change. Compositors may
         not send this event if they do not conceptually arrange workspaces in this way.
+        If compositors simply number workspaces, without any geometric interpretation,
+        they may send 1D coordinates, which clients should not interpret as implying any
+        geometry.
 
         Coordinates have an arbitrary number of dimensions with an uint32 position along
-        each dimension. By convention, the first dimension is X, the second Y, the third
-        Z, and so on. The compositor may chose to utilize these events for a more novel
-        workspace layout convention, however. No guarantee is made about the grid being
-        filled or bounded; there may be a workspace at coordinate 1 and another at
-        coordinate 1000 and none in between.
+        each dimension. By convention if N>1, the first dimension is X, the second Y,
+        the third Z, and so on. The compositor may chose to utilize these events for a
+        more novel workspace layout convention, however. No guarantee is made about the
+        grid being filled or bounded; there may be a workspace at coordinate 1 and
+        another at coordinate 1000 and none in between.
 
         Among a set of workspaces that have identical sets of outputs, workspaces must
         have unique coordinates of the same dimensionality. Coordinates otherwise need
@@ -199,14 +202,14 @@
       </description>
     </request>
 
-    <request name="done">
+    <request name="commit">
       <description summary="all requests about the workspace have been sent">
         The client must send this request after it has finished sending other requests.
-        The compositor must process a series of requests preceding a done request
+        The compositor must process a series of requests preceding a commit request
         atomically. This includes requests on other zwlr_workspace_handle_v1 objects: if
         the client sends requests for multiple zwlr_workspace_handle_v1 objects, the
-        compositor must wait until a done request has been sent for every
-        zwlr_workspace_handle_v1 involved, and process all of the preceding requests
+        compositor must wait until a commit request has been sent for every one of those
+        zwlr_workspace_handle_v1 objects, and process all of the preceding requests
         atomically. This is important so that say, deactivating one workspace and
         activating another can be treated atomically.
 
--- a/unstable/wlr-workspace-unstable-v1.xml
+++ b/unstable/wlr-workspace-unstable-v1.xml
@@ -29,15 +29,16 @@
     <description summary="list and control workspaces">
       Workspaces, also called virtual desktops, are groups of surfaces. A
       compositor with a concept of workspaces may only show some such groups of
-      surfaces (those of 'active' workspaces) at a time. Workspaces may contain
-      surfaces from one or more outputs, and 'activating' a workspace is a
-      request that the compositor display those surfaces on those outputs as
-      normal, whereas the compositor may hide or otherwise de-emphasise surfaces
-      that are associated only with 'inactive' workspaces. By associating a set
-      of outputs with each workspace, each output may have its own unique set of
-      workspaces, or all outputs (or any other arbitrary grouping) may share
-      workspaces. Compositors may optionally conceptually arrange workspaces
-      with a common set of outputs in an N-dimensional grid.
+      surfaces (those of 'active' workspaces) at a time. 'Activating' a
+      workspace is a request for the compositor to display that workspace's
+      surfaces as normal, whereas the compositor may hide or otherwise
+      de-emphasise surfaces that are associated only with 'inactive' workspaces.
+      Workspaces are grouped by which sets of outputs they correspond to, and
+      may contain surfaces only from those outputs. In this way, it is possible
+      for each output to have its own set of workspaces, or for all outputs (or
+      any other arbitrary grouping) to share workspaces. Compositors may
+      optionally conceptually arrange each group of workspaces in an
+      N-dimensional grid.
 
       The purpose of this protocol is to enable the creation of taskbars and
       docks by providing them with a list of workspaces and their properties,
@@ -47,15 +48,15 @@
       sent via the workspace event.
     </description>
 
-    <event name="workspace">
-      <description summary="a workspace has been created">
-        This event is emitted whenever a new workspace has been created.
+    <event name="workspace_group">
+      <description summary="a workspace group has been created">
+        This event is emitted whenever a new workspace group has been created.
 
-        All initial details of the workspace (outputs, name, coordinates) will
-        be sent immediately after this event via the corresponding events in
-        zwlr_workspace_handle_v1.
+        All initial details of the workspace group (workspaces, outputs) will be
+        sent immediately after this event via the corresponding events in
+        zwlr_workspace_group_handle_v1.
       </description>
-      <arg name="workspace" type="new_id" interface="zwlr_workspace_handle_v1"/>
+      <arg name="workspace_group" type="new_id" interface="zwlr_workspace_group_handle_v1"/>
     </event>
 
     <event name="finished">
@@ -70,12 +71,64 @@
     <request name="stop">
       <description summary="stop sending events">
         Indicates the client no longer wishes to receive events for new
-        workspaces. However the compositor may emit further workspace events,
-        until the finished event is emitted.
+        workspace groups. However the compositor may emit further workspace
+        events, until the finished event is emitted.
 
         The client must not send any more requests after this one.
       </description>
     </request>
+  </interface>
+
+  <interface name="zwlr_workspace_group_handle_v1" version="1">
+    <description summary="a workspace group assigned to a set of outputs">
+      A zwlr_workspace_group_handle_v1 object represents a a workspace group
+      that is assigned a set of outputs and contains a number of workspaces.
+
+      The set of outputs assigned to the workspace group is conveyed to the
+      client via enter and leave events, and ita workspaces are conveyed with
+      the workspace event.
+    </description>
+
+    <event name="enter">
+      <description summary="output assigned to workspace group">
+        This event is emitted whenever an output is assigned to the workspace
+        group.
+      </description>
+      <arg name="output" type="object" interface="wl_output"/>
+    </event>
+
+    <event name="leave">
+      <description summary="output removed from workspace group">
+        This event is emitted whenever an output is removed from the workspace
+        group.
+      </description>
+      <arg name="output" type="object" interface="wl_output"/>
+    </event>
+
+    <event name="workspace">
+      <description summary="workspace added to workspace group">
+        This event is emitted whenever a new workspace has been created.
+
+        All initial details of the workspace (name, coordinates, state) will
+        be sent immediately after this event via the corresponding events in
+        zwlr_workspace_handle_v1.
+      </description>
+      <arg name="workspace" type="new_id" interface="zwlr_workspace_handle_v1"/>
+    </event>
+
+    <event name="done">
+      <description summary="all information about the workspace group has been sent">
+        This event is sent after all changes in the workspace group state have
+        been sent.
+
+        This allows changes to one or more zwlr_workspace_group_handle_v1
+        properties to be seen as atomic, even if they happen via multiple
+        events. In particular, an output moving from one workspace group to
+        another sends an enter event and a leave event to the two
+        zwlr_workspace_group_handle_v1 objects in question. The compositor sends
+        both done events only after both other events.
+      </description>
+    </event>
   </interface>
 
   <interface name="zwlr_workspace_handle_v1" version="1">
@@ -84,25 +137,11 @@
       group of surfaces.
 
       Each workspace has a name, conveyed to the client with the name event; a
-      list of states, conveyed to the client with the state event; a list of
-      outputs, conveyed via one or more output events; and optionally a set of
-      coordinates, conveyed to the client with the coordinates event. The client
-      may request that the compositor activate or deactivate the workspace.
+      list of states, conveyed to the client with the state event; and
+      optionally a set of coordinates, conveyed to the client with the
+      coordinates event. The client may request that the compositor activate or
+      deactivate the workspace.
     </description>
-
-    <event name="enter">
-      <description summary="output added to workspace">
-        This event is emitted whenever an output is added to the workspace.
-      </description>
-      <arg name="output" type="object" interface="wl_output"/>
-    </event>
-
-    <event name="leave">
-      <description summary="output removed from workspace">
-        This event is emitted whenever an output is removed from the workspace.
-      </description>
-      <arg name="output" type="object" interface="wl_output"/>
-    </event>
 
     <event name="name">
       <description summary="workspace name changed">
@@ -114,27 +153,23 @@
 
     <event name="coordinates">
       <description summary="workspace coordinates changed">
-        This event is used to organize workspaces spatially into an
-        N-dimensional grid, and if supported, is emitted immediately after the
-        zwlr_workspace_handle_v1 is created and whenever the coordinates of the
-        workspace change. Compositors may not send this event if they do not
+        This event is used to organize workspaces into an N-dimensional grid
+        within a workspace group, and if supported, is emitted immediately after
+        the zwlr_workspace_handle_v1 is created and whenever the coordinates of
+        the workspace change. Compositors may not send this event if they do not
         conceptually arrange workspaces in this way. If compositors simply
         number workspaces, without any geometric interpretation, they may send
         1D coordinates, which clients should not interpret as implying any
         geometry.
 
-        Coordinates have an arbitrary number of dimensions with an uint32
-        position along each dimension. By convention if N>1, the first dimension
-        is X, the second Y, the third Z, and so on. The compositor may chose to
-        utilize these events for a more novel workspace layout convention,
-        however. No guarantee is made about the grid being filled or bounded;
-        there may be a workspace at coordinate 1 and another at coordinate 1000
-        and none in between.
-
-        Among a set of workspaces that have identical sets of outputs,
-        workspaces must have unique coordinates of the same dimensionality.
-        Coordinates otherwise need not be unique: in this way independent grids
-        can exist for each set of outputs that share workspaces.
+        Coordinates have an arbitrary number of dimensions N with an uint32
+        position along each dimension. By convention if N > 1, the first
+        dimension is X, the second Y, the third Z, and so on. The compositor may
+        chose to utilize these events for a more novel workspace layout
+        convention, however. No guarantee is made about the grid being filled or
+        bounded; there may be a workspace at coordinate 1 and another at
+        coordinate 1000 and none in between. Within a workspace group, however,
+        workspaces must have unique coordinates of equal dimensionality.
       </description>
       <arg name="coordinates" type="array"/>
     </event>
@@ -195,8 +230,8 @@
 
         There is no guarantee the workspace will be actually activated, and
         behaviour may be compositor-dependent. For example, activating a
-        workspace may or may not deactivate all other workspaces on the same
-        outputs.
+        workspace may or may not deactivate all other workspaces in the same
+        group.
       </description>
     </request>
 
@@ -217,12 +252,12 @@
         multiple zwlr_workspace_handle_v1 objects, the compositor must wait
         until a commit request has been sent for every one of those
         zwlr_workspace_handle_v1 objects, and process all of the preceding
-        requests atomically. This is important so that say, deactivating one
-        workspace and activating another can be treated atomically.
+        requests atomically. 
 
         This allows changes to the workspace properties to be seen as atomic,
         even if they happen via multiple events, and even if they involve
-        multiple zwlr_workspace_handle_v1 objects.
+        multiple zwlr_workspace_handle_v1 objects, for example, deactivating one
+        workspace and activating another.
       </description>
     </request>
 
--- a/unstable/wlr-workspace-unstable-v1.xml
+++ b/unstable/wlr-workspace-unstable-v1.xml
@@ -198,5 +198,23 @@
         There is no guarantee the workspace will be actually deactivated.
       </description>
     </request>
+
+    <request name="done">
+      <description summary="all requests about the workspace have been sent">
+        The client must send this request after it has finished sending other requests.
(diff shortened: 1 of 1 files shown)


ammen99:
Thanks for starting work on this one!

> This is a draft for addressing issue #10.
> 
> I likely don't have the C chops to contribute much to wlroots' code (though I might try), but maybe I can talk about protocols.
> 
> To accommodate both compositors that have workspaces spanning all outputs, as well as those that have each output with its own independent set of workspaces, this protocol has the concept of 'workspace groups'. A workspace group is a set of outputs that share workspaces. Each output may belong to at most one group. Each workspace group has one active workspace at any one time. This allows arbitrary grouping of outputs, including the two common cases of all outputs sharing a group and of each output being its own group.

I wonder, why do we need workspace groups at all? It seems to me that if we had just workspaces and outputs, where each workspace might be visible on more than one output, that would be enough?
> 
> There is more to think about here.
> 
> Is the protocol sufficient for required atomicity? If an output moves to a different workspace group, then the 'each output may belong to at most one group' requirement will be violated from the perspective of the client until both an `output` and an `output_remove` event are processed on the two `zwlr_workspace_group_handle_v1` objects. But the compositor can refrain from sending `done` events until both of these events are sent, so we could say that if the client gets an `output` or `output_remove` event, it must wait for the `done` event before it has a consistent state for what groups all the outputs belong to.
> 
> Similarly with `zwlr_workspace_handle_v1`, if a workspace changes what its number is, the client must wait for the `done` event in order to see that the 'workspaces are sequentially numbered from zero' requirement is satisfied.
> 
> I am also a bit confused about destroy events and requests. If you request to destroy an object, you're just asking to clean up resources - you're not asking the compositor to remove a workspace or whatnot. However, if the compositor _does_ remove a workspace (not just move it to a different workspace group), does it send a `destroyed` event?

Such a `destroyed` event should be explicitly specified in the protocol. I'm not sure that this is a very good name though, although I can't think of a better one right now.

> 
> I suspect the general ideas in this protocol are pretty unobjectionable, but potentially more contentious things that I haven't included are:
> 
> Workspaces are simply numbered here. Should it be left that way, or should they be defined on a 2D grid? If we do not allow for information about a 2D grid, then if the compositor is laying out the workspaces this way, panels will not be able to reflect that in how they display workspaces, and they will not be able to implement 'move to workspace left' and 'move to workspace up' functionality, etc.
> 
Many window managers don't arrange workspaces in a grid (think for example tagging WMs). If we want this cross-compositor, I think it is best to leave "ordering" of workspaces.

> Should workspaces and/or workspace groups have names?
> 
I am +1 on this, I see no problem in having an optional `name` event so that the compositor can say if the workspace has a name (for ex. Openbox-like) or having them numbered `1,2,...` (Sway) or perhaps something like `1-1, 1-2, ...` for Wayfire.

> Here are some events/requests I have not included
> 
> zwlr_workspace_manager_v1:
> 
> * request: create workspace group
> 
> zwlr_workspace_group_v1:
> 
> * request: add an output to the group
> * request: remove an output from the group
> * request: add a workspace to the group
> * request: remove a workspace from the group
> * request: set a name as a string
> * event: name as a string
> 
> zwl_workspace_v1:
> 
> * event: name as a string
> * request: move (change the number of the workspace - processed as an insert to keep numbers sequential)

Do pagers and the like actually have the ability to reorder workspaces? It seems to me that if this gets included at all, it should be optional, so that the compositor may ignore this request.

> * request: set name as a string
> 
> There is also some question of whether only one workspace can be active at a time. In compositor workspace overview interfaces, you can often see multiple workspaces. However, that's not inconsistent with only one of them being 'active', active doesn't have to mean visible, so I'm not sure if that's a problem.

Thinking of the tagging paradigm I mentioned earlier, I think it may well be that more than a single "workspace" can be active at a given time.

chrisjbillington:
> I wonder, why do we need workspace groups at all? It seems to me that if we had just workspaces and outputs, where each workspace might be visible on more than one output, that would be enough?

It's a good point. Though then, when an output is added or removed, all workspace objects will get events about it, instead of just one workspace_group object. So it will be a bit more chatty, but that's probably fine.

Also, without explicit workspace_group objects, if ever a compositor has a concept of workspaces that has some outputs sharing workspaces  but not others, then a client would have to do some set arithmetic on the sets of outputs from each workspace in order to work out what the groups are. Again, that's probably fine.

So I think I like it, I'll get rid of the workspace_group interface.

> Such a `destroyed` event should be explicitly specified in the protocol. I'm not sure that this is a very good name though, although I can't think of a better one right now.

I'll have to read up on this a bit. I notice in the foreign toplevel management protocol, there is an event `closed` that appears to be notifying the client that the toplevel object has been destroyed. Now, is this event emitted both in the case of the toplevel being closed for some external reason, *and* after the client makes a `destroy` request? If so, it seems like it is playing two roles: one is about resource freeing, the other about the toplevel closing. Closing implies resource freeing, but not necessarily the other way around. So if I understand correctly, I should make separate `workspace_remove` events and `destroy` events, and a `workspace_remove` event will always be followed by a destroy event, but not necessarily the other way around if a `destroy` event is in response to a `destroy` request. And perhaps the foreign toplevel management protocol should distinguish between close and destroy as well. Am I understanding correctly?

> Many window managers don't arrange workspaces in a grid (think for example tagging WMs). If we want this cross-compositor, I think it is best to leave "ordering" of workspaces.

Ok, sounds good.

> I am +1 on this, I see no problem in having an optional `name` event so that the compositor can say if the workspace has a name (for ex. Openbox-like) or having them numbered `1,2,...` (Sway) or perhaps something like `1-1, 1-2, ...` for Wayfire.

Great. Ah, I see grid information could be put in there: clever clients could then still take that into account if they wanted, albeit in a compositor-specific way.

> Do pagers and the like actually have the ability to reorder workspaces? It seems to me that if this gets included at all, it should be optional, so that the compositor may ignore this request.

In gnome-shell at least, if you go into the overview and drag a window into the gaps between workspaces, a new workspace is created there. So that would be equivalent to a 'create' and then a 'move'. But that's all I can think of.

> Thinking of the tagging paradigm I mentioned earlier, I think it may well be that more than a single "workspace" can be active at a given time.

Ok, so if we don't impose that only one workspace can be active per output, then we will need a `deactivate` request as well, since activating another workspace might not imply deactivating all others.

Futher, if there is more than one request for manipulating workspaces, does there need to be `done` requests (or however convention is to name them), so that a set of requests can be treated atomically by the server?

ddevault:
>Great. Ah, I see grid information could be put in there: clever clients could then still take that into account if they wanted, albeit in a compositor-specific way.

Hm, this makes me uncomfortable. I'm not opposed to covering spatial information in this protocol, it just needs to support an arbitrary number of dimensions.

ddevault:
Note that including the workspace name is a good idea, though.

unstable/wlr-workspace-unstable-v1.xml
```diff
@@ -0,0 +1,226 @@
...
+    </event>
+
+    <event name="state">
+      <description summary="the state of the workspace changed">
+        This event is emitted immediately after the zwlr_workspace_handle_v1 is created
+        and each time the workspace state changes, either because of a compositor action
+        or because of a request in this protocol.
+      </description>
+      <arg name="state" type="array"/>
+    </event>
```
ddevault: How does workspace order change atomically?

ddevault: e.g. how to updates affecting multiple workspaces occur atomically

ddevault:
Thought which is probably only going to be useful for future work: some workspace switchers display a little rectangle or something representing where each toplevel is shown on the workspace.

unstable/wlr-workspace-unstable-v1.xml
```diff
@@ -0,0 +1,226 @@
...
+
+    <enum name="state">
+      <description summary="types of states on the workspace">
+        The different states that a workspace can have.
+      </description>
+
+      <entry name="active" value="0" summary="the workspace is active"/>
```
ddevault: The overloading of "active" here is weird. It seems like "active" means some combination of "visible on its output" and "contains toplevels".

ddevault: Maybe this should be a bitfield?

unstable/wlr-workspace-unstable-v1.xml
chrisjbillington: Thanks for reviewing!

Two types of atomicity: to make sure the client knows when the re-ordering is done, my thinking is the compositor refrains from sending a `done` event  until it has sent all the `number` events to the workspace objects whose numbers would have to change as a result of moving one workspace.

The other type of atomicity, where the client doesn't want the compositor to act until it's done sending a bunch of events, I haven't thought of yet. I will need to add a request to say 'requests are done now!', I'll just check how these requests are conventionally named in other protocols.

unstable/wlr-workspace-unstable-v1.xml
chrisjbillington: the state event gives an array of states, so they are not exclusive. I'm copying how toplevel states are conveyed in the foreign toplevel management protocol.

ammen99: I think the protocol just needs to clarify what is "active"

unstable/wlr-workspace-unstable-v1.xml
chrisjbillington: (Though since there are currently no requests in the protocol to move workspaces, it's not necessary yet for the second kind).

chrisjbillington:
Ok, I've pushed some changes.

* There is now a `coordinates` event instead of a `number` event.
 
* There is now a `name` event.

* There is now no imposed concept of workspace groups, just outputs associated with workspaces. The description for `coordinates` does impose extra requirements on the coordinates of workspaces that share a common set of outputs (their coordinates must have equal dimension and be unique), this is the only place where the concept of a group of workspaces remains.

* Added a `deactivate` request, since activating one workspace need not necessarily imply deactivating all others.

* I removed the `empty` state. I imagine we will add either to this or the foreign topevel management protocol the ability to find out which toplevels are on which workspaces, and clients can figure out for themselves if a workspace is empty.

* I renamed the `destroy` event to `remove` for when the workspace is removed for some external reason. I think I understand how destruction works now, and that what I've done makes sense. It matches what the foreign toplevel handles have for when the toplevels are destroyed externally. I now understand that if you call the `destroy` request, the server does not send an event in response, but that if the client receives the `remove` event, it should call `destroy` (correct me if I'm wrong).

Still no abilities to manipulate workspaces beyond activating and deactivating them, so we can talk more about that still.

chrisjbillington:
> Thought which is probably only going to be useful for future work: some workspace switchers display a little rectangle or something representing where each toplevel is shown on the workspace.

perhaps the foreign toplevel protocol ought to grow `workspace`, `workspace_remove` and `geometry` events. I'll certainly need the first two to be implemented (somewhere, if not in the foreign toplevel protocol) for my panel to know which toplevels are on which workspaces.

unstable/wlr-workspace-unstable-v1.xml
chrisjbillington: Ok, I've defined a little better in the overall protocol description what 'active' means, hopefully that is satisfactory. Though really it is up to the compositor what it means, so the definition is not super precise.

unstable/wlr-workspace-unstable-v1.xml
```diff
@@ -0,0 +1,192 @@
...
+      surfaces.
+
+      Each workspace has a name, conveyed to the client with the name event; a list of
+      states, conveyed to the client with the state event; a list of outputs, conveyed
+      via one or more output events; and optionally a set of coordinates, conveyed to
+      the client with the coordinates event. The client may request that the compositor
+      activate or deactivate the workspace.
+    </description>
+
+    <event name="output">
```
emersion: I'd name this `enter` (and rename `output_remove` to `leave`) for consistency with the rest of the Wayland protocol.

unstable/wlr-workspace-unstable-v1.xml
```diff
@@ -0,0 +1,192 @@
...
+      </description>
+      <arg name="output" type="object" interface="wl_output"/>
+    </event>
+
+    <event name="name">
+      <description summary="workspace name changed">
+        This event is emitted immediately after the zwlr_workspace_handle_v1 is created
+        and whenever the name of the workspace changes.
+      </description>
+      <arg name="number" type="int"/>
```
emersion: Shouldn't this be a string?

unstable/wlr-workspace-unstable-v1.xml
```diff
@@ -0,0 +1,192 @@
...
+    </event>
+
+    <event name="coordinates">
+      <description summary="workspace coordinates changed">
+        If the compositor assigns coordinates to workspaces, this event is emitted
+        immediately after the zwlr_workspace_handle_v1 is created and whenever the
+        coordinates of the workspace change. Coordinates have an arbitrary number of
+        dimensions with an unsigned integer position for each dimension. Among a set of
```
emersion: Should specify the width of the uint. 32 bits?

unstable/wlr-workspace-unstable-v1.xml
```diff
@@ -0,0 +1,192 @@
...
+        workspaces that have identical sets of outputs, workspaces must have unique
+        coordinates with the same number of dimensions, such that the workspaces form a
+        grid with at most one workspace in each position. By convention, 2D coordinates
+        represent [y_pos, x_pos].
```
emersion: That's unconventional, is there a reason to do this?

chrisjbillington: Row before column is the convention for image pixel coordinates and matrix indexing as far as I'm aware. I might change the text to say [row, column] instead though, since x and y are more often used for continuous variables where the convention is opposite.

emersion: I'd prefer `x, y, z, …` because that way `pos[0]` is always x, `pos[1]` is always y and so on.

I'm still not sure it's a good idea to have this in the protocol.

chrisjbillington: I can just leave it out, just there's a risk that different compositors will choose different conventions, and then clients won't be able to represent them graphically. Happy to go with [column, row] too.

ddevault: What if someone wants to make 3D workspaces, or even 4D? I'd suggest adding an event which has `dimension` and `extent`.

emersion: wl_array already contains the size of the array

ddevault: Derp. I don't like wl_array, I'd rather raise the event several times.

emersion: No need to make things more complicated because you personally don't like wl_array.

ddevault: What I suggest is more type-safe and doesn't involve manually reasoning about the type of an opaque array.

chrisjbillington: About the index ordering, the convention I'm used to is that you would consider a 3D array to be an array of 2D arrays, such that the new index going from 2D to 3D gets inserted at the start. So it would be [z, y, x]. 

How about:
```For the purposes of representing grids of workspaces graphically, for dimension > 1 clients may interpret the final two indices as [row_index, column_index] of the workspace in a two-dimensional grid.```

I'm happy with switching this to "first two indices represent [column, row]" but I do think that is the less common convention. And I'm leaning toward saying the convention should be mentioned in the protocol, otherwise clients will not be able to reliably represent the grids in a way that will match how the compositor is representing them.

Comment (ddevault) [reply unstable/wlr-workspace-unstable-v1.xml]: I think it should be ordered x, y, z, w, ...; rather than with respect to the implementation details.

Comment (chrisjbillington) [reply unstable/wlr-workspace-unstable-v1.xml]: an extent event at least would be necessary if the grid can have empty slots - you cant infer the extent from existing workspaces in any case if there are allowed to be gaps.

The dimensionality change events would simplify client code at the expense of more events in the protocol. I've been leaning away from this (for example removing the `empty` state since this can be inferred from other information), but is there is a precedent in wayland as to whether the protocols should provide redundant information for the sake of simplifying client code? I would have guessed that there is an overall attitude regarding this one way or the other. (and I would have guessed it was *not* to provide redundant info even if it simplifies things for the client).

Comment (ddevault) [reply unstable/wlr-workspace-unstable-v1.xml]: I'm not suggesting adding more events - just a single event with two parameters, dimension and extent, which is raised multiple times in a single atomic transaction.

I'm not sure there's any precedent for protocols attempting to solve a problem as generic as "workspaces".

Comment (chrisjbillington) [reply unstable/wlr-workspace-unstable-v1.xml]: Alright, I'l put an end to the bikeshedding and declare it's [x, y] in the first two indices.

Comment (ddevault) [reply unstable/wlr-workspace-unstable-v1.xml]: That fails to address the concerns about compositors with different dimensionalities.

Comment (chrisjbillington) [reply unstable/wlr-workspace-unstable-v1.xml]: What do you mean it's raised multiple times? Wouldn't it only be raised when the dimension or extent change? I must be missing something.

Comment (chrisjbillington) [reply unstable/wlr-workspace-unstable-v1.xml]: Compositors with higher dimensionalities may use more indices, but the client can interpret the first two as [x, y]. So for a 3D array, the client can interpret the space as a stack of 2D grids, with the first two indices being [x, y] and the third index being which grid is displayed at any time (call this z if you like).

But I could change the language to be more generic: "if representing graphically, clients may interpret coords as x, y, z, etc in a spatial grid" or some such, rather than only mentioning [x, y] two.

Comment (ddevault) [reply unstable/wlr-workspace-unstable-v1.xml]: But where does the third index come from in the protocol? Also, I prefer this language:

>By convention this event is used to organize workspaces spatially into an N-dimensional grid. By convention, the first dimension is X, the second Y, the third Z, and so on. The compositor may chose to utilize these events for a more novel workspace layout convention, however. No guarantee is made about the grid being filled or bounded; there may be a workspace at coordinate 1 and another at coordinate 1000 and none in between.

Comment (chrisjbillington) [reply unstable/wlr-workspace-unstable-v1.xml]: And compositors with lower dimensionality don't need a convention - if they want to specify that their 1D list of workspaces should be considered horizontally or vertically arranged, they can use 2D coords with a single row or a single column. If they don't care and it's just "next/previous" without a geometric interpretation then they can just give a 1D index.

Comment (ddevault) [reply unstable/wlr-workspace-unstable-v1.xml]: Forcing 1D compositors to contort themselves into two dimensions is a bad protocol. That doesn't consider 3D compositors, either.

Comment (chrisjbillington) [reply unstable/wlr-workspace-unstable-v1.xml]: I like your proposed language, but I don't see how we're disagreeing, i thought that was what I was saying already. I'll just adopt your language and then I suppose we'll be happy on that point.

Comment (chrisjbillington) [reply unstable/wlr-workspace-unstable-v1.xml]: If a compositor cares about whether it's workspaces are represented horizontally or vertically, then it is 2D in a sense, even if there is only one row or column. Leaving this info in the coordinates seems like a better protocol than having separate events telling you how to interpret 1D coordinates geometrically. But this comes again to whether it's a good idea to have redundancy for the sake of convenience.

Comment (chrisjbillington) [reply unstable/wlr-workspace-unstable-v1.xml]: Oh I see what you were saying. By only specifying the convention for 2D and higher, the meaning of a 1D coordinate is unspecified. Whereas with your wording, a 1D coordinate should be interpreted as a horizontal coordinate. So gnome-shell will need to give 2D coordinates to communicate that its workspaces are vertical, but other compositors can use 1D coords for their horizontal workspaces.

Comment (ddevault) [reply unstable/wlr-workspace-unstable-v1.xml]: gnome-shell should do no such thing - clients should let the user specify how to arrange their workspaces, vertically or horizontally, in 1D. X/Y isn't meaningful in 1 dimension, so the X coordinate being arranged vertically isn't a problem. When transposing between coordinate systems you can losslessly transpose one coordinate onto "another" - the X coordinate in 2D doesn't mean the same thing as the X coordinate in 1D.

**Comment by chrisjbillington** [comment]
What do you propose gnome-shell send to clients to tell them that if they want to be consistent with how gnome-shell represents the workspaces in its overview interface, that they should represent the workspaces vertically?

Even if this was set by a client, it needs to be communicated to subsequent clients. Sending 2D coords to communicate this seems to me like a fine way to do it.

File: unstable/wlr-workspace-unstable-v1.xml (hunk 6 above)
Comment (chrisjbillington) [reply unstable/wlr-workspace-unstable-v1.xml]: What do you propose gnome-shell send to clients to tell them that if they want to be consistent with how gnome-shell represents the workspaces in its overview interface, that they should represent the workspaces vertically?

Even if this was set by a client, it needs to be communicated to subsequent clients. Sending 2D coords to communicate this seems to me like a legitimate and unproblematic way to do it.

Comment (ddevault) [reply unstable/wlr-workspace-unstable-v1.xml]: >What do you propose gnome-shell send to clients to tell them that if they want to be consistent with how gnome-shell represents the workspaces in its overview interface, that they should represent the workspaces vertically?

I would suggest that it doesn't. Besides - what if the user wants their workspaces shown horizontally in their panel and vertically in GNOME? We shouldn't add a dimension that doesn't exist.

Comment (chrisjbillington) [reply unstable/wlr-workspace-unstable-v1.xml]: That's acceptable to me, but would be a bit strange because you could click on a workspace oriented to the right of the current one in your panel app, and gnome-shell would show a vertical animation. Windows overlapping multiple workspaces would display strangely if the panel app showed their positions with squares.

I'd argue the dimension exists because of these animations and window overlapping. But I'm happy to leave it unspecified and leave it up to gnome-shell to decide what to send - it can send a 2d coord if it wants.

Comment (ammen99) [reply unstable/wlr-workspace-unstable-v1.xml]: guys, I really think we should let it be in some free form like a string. There are many different combinations that are possible, and that one here doesn't address tagging wm's.

Comment (emersion) [reply unstable/wlr-workspace-unstable-v1.xml]: >guys, I really think we should let it be in some free form like a string. There are many different combinations that are possible, and that one here doesn't address tagging wm's.

+1. I wouldn't address coords.

It should be enough to get the output layout using xdg-output, and use the workspace::enter,leave events to know non which output(s) a workspace is on.

Comment (ddevault) [reply unstable/wlr-workspace-unstable-v1.xml]: That would work for sway, but not for GNOME, for which workspaces are arranged at a higher level than outputs.

Comment (ammen99) [reply unstable/wlr-workspace-unstable-v1.xml]: as I said, we can just make it a string and say that compositors arranging in a grid would follow `x-y-z-...` format or whatever we find convenient. This way we also leave space for compositors that do not arrange in a grid, and they can then use their own specific way to encode data using that field in the protocol. IMO there is no one-size-fits-all solution.

Comment (ammen99) [reply unstable/wlr-workspace-unstable-v1.xml]: plus, in Wayfire I can't really make workspaces arranged in a grid in multimonitor configuration, they aren't in the same coordinate system.

Comment (chrisjbillington) [reply unstable/wlr-workspace-unstable-v1.xml]: The name field works for tagging, no?

Coordinates seem common enough that if we leave it free-form, I imagine an ad-hoc standard for making strings of coordinates will present itself (and that gnome would end up sending 2D coordinates within the strings to tell clients that its workspaces are vertical). So by making it a string we're just deferring that standard to be a de-facto one instead of in the protocol. The benefit is allowing information about non-grid-like layouts.

But if it's an opaque string, what's it called? 'layout?' 'geometry'?

My preferred outcome is coordinates with strictly defined interpretation as x, y, z, etc, with compositors without such a concept just not using this part of the protocol. We can add extra events for other paradigms, including a catch-all opaque string for compositor-specific info.

Comment (ammen99) [reply unstable/wlr-workspace-unstable-v1.xml]: hm okay, if we make this event optional, I'm ok

Comment (chrisjbillington) [reply unstable/wlr-workspace-unstable-v1.xml]: > plus, in Wayfire I can't really make workspaces arranged in a grid in multimonitor configuration, they aren't in the same coordinate system.

The coordinates are only meaningful among workspaces that share a set of outputs. So you can have a workspace at (0,0) on output 1, and a workspace on (0,0) on output 2. Or you can have a workspace at (0,0) that is attached to both outputs 1 and 2. But you cannot have *two* workspaces with coords (0,0) on the same outputs. This all may have been more clear with the original concept of 'workspace groups' - that the coordinates are only meaningful within a  workspace group that share a common set of outputs.

Comment (ddevault) [reply unstable/wlr-workspace-unstable-v1.xml]: >an ad-hoc standard for making strings of coordinates will present itself

This sounds *fucking awful*.

**Comment by ddevault** [comment]
Let's move the coordinates discussion out into the main thread. We need a design which accomodates the needs of all compositors in an unambiguous and type-safe manner. This includes the workspace designs of sway, waymonad, KDE, and GNOME.

File: unstable/wlr-workspace-unstable-v1.xml (hunk 6 above)
Comment (chrisjbillington) [reply unstable/wlr-workspace-unstable-v1.xml]: > This sounds _fucking awful_.

:+1:

**Comment by chrisjbillington** [comment]
I'll fix the issues unrelated to coordinates, and modify the coordinate stuff to what I think the best result of the discussion we've had so far on coordinates has been (admittedly opinionated), and will revisit after said discussion.

**Comment by chrisjbillington** [comment]
I've added a `done` request so that atomic requests can be made, for example, activating one workspace and deactivating another.

Firstly: no problem with a request having the same name as an event?

Secondly: I can't find similar requests in other wayland protocols. Is there a reason for this? Are batches of client requests intrinsically treated as atomic by the protocol such that `done` requests are unnecessary?

**Comment by emersion** [comment]
>Firstly: no problem with a request having the same name as an event?

This is a bad idea, too confusing.

`done` is when the compositor wants to atomically send a state. Other names (e.g. `commit`) are used when the client atomically sends a state.

**Comment by chrisjbillington** [comment]
Excellent. `commit` it is then.

**Comment by chrisjbillington** [comment]
Hi all,

How do we feel about the current state of this for a v1? See also [my comment in the issue thread](https://github.com/swaywm/wlr-protocols/issues/10#issuecomment-468020681) explaining the "workspace group" concept and other terminology.

**Comment by ddevault** [comment]
Do we have an example client & server implementation yet?

**Comment by AdrianVovk** [comment]
@ammen99 thoughts about this? I could implement this with carbonSHELL's overview mode. With some layer-shell changes, the only wayfire-specific proto I'd depend on will be some way to trigger my switcher plugin (in theory).

**Comment by ammen99** [comment]
Is there somebody working on an implementation for this in wlroots? I'd pick it up if there are no current ongoing efforts.

File: unstable/wlr-workspace-unstable-v1.xml [hunk 7]
```diff
@@ -0,0 +1,265 @@
...
+        position along each dimension. By convention if N > 1, the first
+        dimension is X, the second Y, the third Z, and so on. The compositor may
+        chose to utilize these events for a more novel workspace layout
+        convention, however. No guarantee is made about the grid being filled or
+        bounded; there may be a workspace at coordinate 1 and another at
+        coordinate 1000 and none in between. Within a workspace group, however,
+        workspaces must have unique coordinates of equal dimensionality.
+      </description>
+      <arg name="coordinates" type="array"/>
+    </event>
```
ddevault: This does not define the dimensionality, be it 1D, 2D, etc. We need a second argument for that.

ddevault:
@ammen99 I don't think anyone is actively working on this, feel free.

unstable/wlr-workspace-unstable-v1.xml
chrisjbillington: I considered the length of the array to be the dimensionality. Is that insufficient? (or non-obvious enough that I should mention it explicitly)?

emersion: It should be fine, xdg-shell already does the same.

unstable/wlr-workspace-unstable-v1.xml
```diff
@@ -0,0 +1,265 @@
...
+
+    <request name="deactivate">
+      <description summary="activate the workspace">
+        Request that this workspace be deactivated.
+
+        There is no guarantee the workspace will be actually deactivated.
+      </description>
+    </request>
+
+    <request name="commit">
```
ammen99, line 246: Why should that be here? Can't we move commit to the manager interface, since this request seems to synchronize everything?

emersion: Indeed, this would make more sense.

ammen99:
One more thing, I noticed that we have many done events. Shouldn't we like have a single done in the manager, just as I proposed to have a single one for the commit request?

Otherwise, I think that depending on interpretation, we can have:

1. Done in workspace group is related only to which outputs this group belongs, and which workspaces are part of it. Then, I don't think we have a way to atomically change activated workspace from one to the other, do we?

2. Done in workspace group applies to also workspaces in this group. Then, my question would be, what is the point of done in each individual workspace?

coderkalyan:
> Thought which is probably only going to be useful for future work: some workspace switchers display a little rectangle or something representing where each toplevel is shown on the workspace.

For my compositor's future workspace switch panel, I want to be able to accurately display the contents of each workspace, with an image instead of just a few rectangles. I'm not sure what would be the best way to implement this, but could we have some kind of system so the client can request the compositor to return the most recent rendered buffer from a workspace? Sort of like screencopy, but where the last composited frame of a workspace is saved.

For sake of argument, this is what the macOS virtual desktop switcher displays:
image

ammen99:
@coderkalyan this looks like something that should be part of the compositor and not a client?

unstable/wlr-workspace-unstable-v1.xml
chrisjbillington: OK, I'll make this change.

coderkalyan:
@ammen99 I'm just talking about the top part of that image - the desktop switcher just shows a workspace-shot instead of a screen-shot :)

ifreund:
I've been working on river which is a tagging compositor in the same vein as dwm, xmonad, etc. I thought that it would be good to mention that I've created a [river-status](https://github.com/ifreund/river/blob/master/protocol/river-status-unstable-v1.xml) which river uses to communicate this information to status bars (specifically the `output-status` interface).  There's also a client implementation of part of this protocol in waybar [here](https://github.com/Alexays/Waybar/blob/master/src/modules/river/tags.cpp).

One thing that I haven't seen clarified yet is if a given surface can belong to multiple workspaces? This is the case with tags in river: a single view may belong to an arbitrary set of tags, and an arbitrary set of tags may be displayed on a given output at any given time. Both tags and views are local to the output.

As stated earlier, workspace coordinates make absolutely no sense for a tagging compositor like river.

Overall though, this protocol looks promising and I would be interested in implementing it in river given that surfaces may infact belong to multiple outputs (which should be mentioned in the protocol) and that coordinates remain optional or are removed.

ammen99:
> Overall though, this protocol looks promising and I would be interested in implementing it in river given that surfaces may infact belong to multiple workspaces (which should be mentioned in the protocol) and that coordinates remain optional or are removed.

My understanding is that the link between workspace-unstable and foreign-toplevel-management-unstable is not yet done, and will probably be an event in foreign-toplevel-management (but I 100% agree that a toplevel can be on more than one workspace, this is the case in Wayfire too).

ammen99:
wayland-protocols MR: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/40

emersion:

https://gitlab.freedesktop.org/wlroots/wlr-protocols/-/merge_requests/35

PR #38, opened 2019-03-07
Add wlr-output-management-unstable-v1

emersion:
This commit adds a new output management protocol. It can be used to get the
current output configuration and apply a new one. It's based on the GNOME D-Bus
API [1] and the KDE protocol [2] [3].

Goals:

* High-level API, not a copy of KMS
* Protect against races wrt. hotplugging
* Extensible via backwards-compatible changes
* Atomic (both when reading current config and when applying a new one)

Features:

* Read all outputs including disabled ones
  * Identification (name, description), available modes, physical size
  * Current mode, enabled/disabled
  * Position, scale, transform
* Write
  * Modeset, enable/disable
  * Position, scale, transform

Non-features (ie. features in GNOME or KDE excluded from this protocol):

* No DPMS (left out for another protocol because it can operate on wl_output)
* No gamma ramps (left out for another protocol, we already have one)
* Cannot set the primary logical monitor and the presentation monitor (should be
  a separate protocol, could use wl_output)

Available outputs and their current configuration are advertised when the client
binds to the manager as "heads" (a head is an enabled or disabled output).

Clients can create a configuration object, add some outputs to configure and
apply the new state.

The GNOME protocol has a concept of "logical monitor": it's a region of the
screen you can assign physical monitors to. You can modeset and enable/disable
physical monitors while you can position, scale and transform logical monitors.
This nicely splits physical properties and logical properties. I've tried to
design a protocol with a similar concept [4], but it turned out being pretty
complicated since you need to have at least two extra objects (one to read the
current logical monitors, one to configure them). I've decided against this
approach for now.

TODO:

- [x] Custom modes
- [x] Check all of this makes sense for non-KMS outputs?
  - Some outputs have fixed modes, a size, a fixed refresh rate (KMS)
  - Some outputs have fixed modes, a size, a variable refresh rate (KMS + VRR)
  - Some outputs have no modes, a size, no refresh rate (Wayland backend)
  - Some outputs have no modes, a size, a refresh rate (headless backend)
- [x] Make sure a compositor can offload everything to a helper client (ie. it
  doesn't apply an initial default modeset on hotplug)
- [x] Should some features be optional?
- [x] Should zero modes mean the output doesn't support modes? What to do if an EDID
  is malformed and returns zero modes?

Possible future features:

* Modes aspect ratio and interlaced modes
* Underscanning? (needs investigation)
* is-builtin property for heads? (should this be in the compositor?)
* Expose suggested panel orientation and tiling data?
* Variable refresh rate? (needs investigation)

Let me know what you think!

Fixes https://github.com/swaywm/wlr-protocols/issues/15

[1]: https://gitlab.gnome.org/GNOME/mutter/blob/master/src/org.gnome.Mutter.DisplayConfig.xml#L292
[2]: https://github.com/KDE/kwayland/blob/master/src/client/protocols/outputdevice.xml
[3]: https://github.com/KDE/kwayland/blob/master/src/client/protocols/output-management.xml
[4]: https://paste.sr.ht/%7Eemersion/8727ea50437fff3927c81c40c0ecc9c9885817ef

### DIFF
--- a/unstable/wlr-output-management-unstable-v1.xml
+++ b/unstable/wlr-output-management-unstable-v1.xml
@@ -292,7 +292,8 @@
 
       First, the client needs to setup the output configuration. Each head can
       be either enabled (and configured) or disabled. It is a protocol error to
-      send two enable_head or disable_head requests with the same head.
+      send two enable_head or disable_head requests with the same head. It is a
+      protocol error to omit a head in a configuration.
 
       Then, the client can apply or test the configuration. The compositor will
       then reply with a succeeded, failed or cancelled event. Finally the client
@@ -300,9 +301,11 @@
     </description>
 
     <enum name="error">
-      <entry name="head_already_configured" value="1"
+      <entry name="already_configured_head" value="1"
         summary="head has been configured twice"/>
-      <entry name="already_used" value="2"
+      <entry name="unconfigured_head" value="2"
+        summary="head has not been configured"/>
+      <entry name="already_used" value="3"
         summary="request sent after configuration has been applied or tested"/>
     </enum>
 
--- a/unstable/wlr-output-management-unstable-v1.xml
+++ b/unstable/wlr-output-management-unstable-v1.xml
@@ -185,28 +185,10 @@
       <arg name="height" type="int" summary="height in millimeters of the output"/>
     </event>
 
-    <event name="arbitrary_modes">
-      <description summary="head supports arbitrary modes">
-        This event is sent if the head supports arbitrary modes.
-
-        When arbitrary modes are not supported, the client should prefer to use
-        fixed modes advertised with the mode event.
-
-        When arbitrary modes are supported, there is no fixed list of modes. The
-        only mode is the current mode. Custom modes will need to be used to
-        change the current mode.
-      </description>
-    </event>
-
     <event name="mode">
       <description summary="introduce a mode">
-        This event introduces a mode for this head.
-
-        If the head doesn't support arbitrary modes, this event is sent once
-        per supported mode.
-
-        If the head supports arbitrary modes, this event is sent once for the
-        current mode if the head is enabled.
+        This event introduces a mode for this head. It is sent once per
+        supported mode.
       </description>
       <arg name="mode" type="new_id" interface="zwlr_output_mode_v1"/>
     </event>
@@ -430,13 +412,16 @@
   <interface name="zwlr_output_configuration_head_v1" version="1">
     <description summary="head configuration">
       This object is used by the client to update a single head's configuration.
+
+      It is a protocol error to set the same property twice.
     </description>
 
     <enum name="error">
-      <entry name="invalid_mode" value="1" summary="mode doesn't belong to head"/>
-      <entry name="invalid_custom_mode" value="2" summary="mode is invalid"/>
-      <entry name="invalid_transform" value="3" summary="transform value outside enum"/>
-      <entry name="invalid_scale" value="4" summary="scale negative or zero"/>
+      <entry name="already_set" value="1" summary="property has already been set"/>
+      <entry name="invalid_mode" value="2" summary="mode doesn't belong to head"/>
+      <entry name="invalid_custom_mode" value="3" summary="mode is invalid"/>
+      <entry name="invalid_transform" value="4" summary="transform value outside enum"/>
+      <entry name="invalid_scale" value="5" summary="scale negative or zero"/>
     </enum>
 
     <request name="set_mode">
@@ -451,6 +436,8 @@
         This request assigns a custom mode to the head. The size is given in
         physical hardware units of the output device. If set to zero, the
         refresh rate is unspecified.
+
+        It is a protocol error to set both a mode and a custom mode.
       </description>
       <arg name="width" type="int" summary="width of the mode in hardware units"/>
       <arg name="height" type="int" summary="height of the mode in hardware units"/>
--- a/unstable/wlr-output-management-unstable-v1.xml
+++ b/unstable/wlr-output-management-unstable-v1.xml
@@ -26,7 +26,7 @@
   </copyright>
 
   <description summary="protocol to configure output devices">
-    This protocol exposes interfaces to get and change output device
+    This protocol exposes interfaces to obtain and modify output device
     configuration.
 
     Warning! The protocol described in this file is experimental and
@@ -116,19 +116,19 @@
 
   <interface name="zwlr_output_head_v1" version="1">
     <description summary="output device">
-      A head is an output device. The difference with wl_output is that heads
-      are advertized even if they are turned off. A head object only advertises
-      properties and cannot be used directly to change them. In order to update
-      some properties, one needs to create a wlr_output_configuration object.
+      A head is an output device. The difference between a wl_output object and
+      a head is that heads are advertised even if they are turned off. A head
+      object only advertises properties and cannot be used directly to change
+      them.
 
       A head has some read-only properties: mode, name, description and
       physical_size. These cannot be changed by clients.
 
-      enabled and current_mode are physical properties. Updating them might take
-      some time, depending on hardware limitations.
+      Other properties can be updated via a wlr_output_configuration object.
 
-      position, transform and scale are logical properties. They describe how
-      the output is mapped in the global compositor space.
+      Properties sent via this interface are applied atomically via the
+      wlr_output_manager.done event. No guarantees are made regarding the order
+      in which properties are sent.
     </description>
 
     <event name="mode">
@@ -168,7 +168,9 @@
 
         The description is a UTF-8 string with no convention defined for its
         contents. Examples might include 'Foocorp 11" Display' or 'Virtual X11
-        output via :1'.
+        output via :1'. However, do not assume that the name is a reflection of
+        the make, model, serial of the underlying DRM connector or the display
+        name of the underlying X11 connection, etc.
 
         If the compositor implements xdg-output and this head is enabled,
         the xdg_output.description must report the same description.
@@ -251,6 +253,10 @@
 
       Some heads don't support output modes, in which case modes won't be
       advertised.
+
+      Properties sent via this interface are applied atomically via the
+      wlr_output_manager.done event. No guarantees are made regarding the order
+      in which properties are sent.
     </description>
 
     <event name="size">
@@ -322,7 +328,7 @@
 
     <request name="disable_head">
       <description summary="disable a head">
-        Disable a head. The head's properties are irrelevant in this case.
+        Disable a head.
       </description>
       <arg name="head" type="object" interface="zwlr_output_head_v1"
         summary="the head to be disabled"/>
--- a/unstable/wlr-output-management-unstable-v1.xml
+++ b/unstable/wlr-output-management-unstable-v1.xml
@@ -46,15 +46,23 @@
 
       Output devices that display pixels (e.g. a physical monitor or a virtual
       output in a window) are represented as heads. Heads cannot be created nor
-      destroyed, but they can be enabled or disabled and their properties can be
-      changed. Each head may have one or more available modes.
+      destroyed by the client, but they can be enabled or disabled and their
+      properties can be changed. Each head may have one or more available modes.
 
-      Heads are advertised when the output manager is bound, and whenever they
-      appear.
+      Whenever a head appears (e.g. a monitor is plugged in), it will be
+      advertised via the head event. Immediately after the output manager is
+      bound, all current heads are advertised.
 
-      Whenever the number of heads or modes changes, the done event will be
-      sent. It carries a serial which can be used in a create_configuration
-      request to change heads properties.
+      Whenever a head's properties change, the relevant wlr_output_head events
+      will be sent. Not all head properties will be sent: only properties that
+      have changed need to.
+
+      Whenever a head disappears (e.g. a monitor is unplugged), a
+      wlr_output_head.finished event will be sent.
+
+      After one or more heads appear, change or disappear, the done event will
+      be sent. It carries a serial which can be used in a create_configuration
+      request to update heads properties.
 
       The information obtained from this protocol should only be used for output
       configuration purposes. This protocol is not designed to be a generic
@@ -64,7 +72,9 @@
 
     <event name="head">
       <description summary="introduce a new head">
-        This event introduces a new head.
+        This event introduces a new head. This happens whenever a new head
+        appears (e.g. a monitor is plugged in) or after the output manager is
+        bound.
       </description>
       <arg name="head" type="new_id" interface="zwlr_output_head_v1"/>
     </event>
@@ -75,7 +85,8 @@
         the output manager object and after any subsequent changes. This applies
         to child head and mode objects as well. In other words, this event is
         sent whenever a head or mode is created or destroyed and whenever one of
-        their properties has been changed.
+        their properties has been changed. Not all state is re-sent each time
+        the current configuration changes: only the actual changes are sent.
 
         This allows changes to the output configuration to be seen as atomic,
         even if they happen via multiple events.
@@ -121,7 +132,7 @@
       object only advertises properties and cannot be used directly to change
       them.
 
-      A head has some read-only properties: mode, name, description and
+      A head has some read-only properties: modes, name, description and
       physical_size. These cannot be changed by clients.
 
       Other properties can be updated via a wlr_output_configuration object.
@@ -371,6 +382,9 @@
         tested them.
 
         Upon receiving this event, the client should destroy this object.
+
+        If the current configuration has changed, events to describe the changes
+        will be sent followed by a wlr_output_manager.done event.
       </description>
     </event>
 
--- a/unstable/wlr-output-management-unstable-v1.xml
+++ b/unstable/wlr-output-management-unstable-v1.xml
@@ -131,13 +131,6 @@
       in which properties are sent.
     </description>
 
-    <event name="mode">
-      <description summary="advertise a supported mode">
-        If the head supports modes, this event is sent once per supported mode.
-      </description>
-      <arg name="mode" type="new_id" interface="zwlr_output_mode_v1"/>
-    </event>
-
     <event name="name">
       <description summary="head name">
         This event describes the head name.
@@ -192,6 +185,32 @@
       <arg name="height" type="int" summary="height in millimeters of the output"/>
     </event>
 
+    <event name="arbitrary_modes">
+      <description summary="head supports arbitrary modes">
+        This event is sent if the head supports arbitrary modes.
+
+        When arbitrary modes are not supported, the client should prefer to use
+        fixed modes advertised with the mode event.
+
+        When arbitrary modes are supported, there is no fixed list of modes. The
+        only mode is the current mode. Custom modes will need to be used to
+        change the current mode.
+      </description>
+    </event>
+
+    <event name="mode">
+      <description summary="introduce a mode">
+        This event introduces a mode for this head.
+
+        If the head doesn't support arbitrary modes, this event is sent once
+        per supported mode.
+
+        If the head supports arbitrary modes, this event is sent once for the
+        current mode if the head is enabled.
+      </description>
+      <arg name="mode" type="new_id" interface="zwlr_output_mode_v1"/>
+    </event>
+
     <event name="enabled">
       <description summary="head is enabled or disabled">
         This event describes whether the head is enabled. A disabled head is not
@@ -206,7 +225,7 @@
     <event name="current_mode">
       <description summary="current mode">
         This event describes the mode currently in use for this head. It is only
-        sent if the output is enabled and supports modes.
+        sent if the output is enabled.
       </description>
       <arg name="mode" type="object" interface="zwlr_output_mode_v1"/>
     </event>
@@ -272,7 +291,8 @@
 
     <event name="refresh">
       <description summary="mode refresh rate">
-        This event describes the mode's fixed vertical refresh rate, if any.
+        This event describes the mode's fixed vertical refresh rate. It is only
+        sent if the mode has a fixed refresh rate.
       </description>
       <arg name="refresh" type="int" summary="vertical refresh rate in mHz"/>
     </event>
@@ -414,8 +434,9 @@
 
     <enum name="error">
       <entry name="invalid_mode" value="1" summary="mode doesn't belong to head"/>
-      <entry name="invalid_transform" value="2" summary="transform value outside enum"/>
-      <entry name="invalid_scale" value="3" summary="scale negative or zero"/>
+      <entry name="invalid_custom_mode" value="2" summary="mode is invalid"/>
+      <entry name="invalid_transform" value="3" summary="transform value outside enum"/>
+      <entry name="invalid_scale" value="4" summary="scale negative or zero"/>
     </enum>
 
     <request name="set_mode">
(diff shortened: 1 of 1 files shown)


ascent12:
Overall the concept seems good to me. I think the `wlr_output` API would need to change to accommodate this cleanly, but this is the type of change I've wanted to make for ages; instead of specifying each output individually, you can do it all in one go atomically (at least as atomically as the backend allows).

To bring up a point about terminology, going by what Weston uses, a "head" is the thing which is displaying pixels (e.g. a computer monitor), and an "output" is the logical representation of that. An output can be assigned multiple heads, which can be implemented efficiently by sharing CRTCs where possible.

This type of efficient cloning is something I'd want to see in a protocol like this and in our APIs, but really does require the split concept of a "logical monitor".

>You can modeset and enable/disable
physical monitors while you can position, scale and transform logical monitors.
This nicely splits physical properties and logical properties. I've tried to
design a protocol with a similar concept, but it turned out being pretty
complicated since you need to have at least two extra objects (one to read the
current logical monitors, one to configure them).

That's not the way I'd go about it. Instead heads would be purely constant/descriptive.
All of the modesetting and property setting would happen on the outputs, and be applied to all heads.
Ignoring the exact names, something along the lines of
- `wlr_output_manager.create_configuration()`
- `wlr_output_manager.create_output()`
- `wlr_output.attach_head()` (call one or more times)
- `wlr_output.set_mode()`
- `wlr_output.set_*()`
- `wlr_output_configuration.attach_output()`
- `wlr_output_configuration.test()`
- `wlr_output_configuration.commit()`
Other things related to that:
- While modes are advertised on heads but set on outputs, a user may wonder from which head they take the mode. Perhaps it shouldn't matter, or we remove the separate mode object and just specify it with width*height@rate, and it's up to the user to make sure it picks a mode suitable to all heads involved.
- If you want to disable a head, create a new configuration which does not include it.

ascent12:
Actually, reading https://paste.sr.ht/%7Eemersion/8727ea50437fff3927c81c40c0ecc9c9885817ef a little more closely, it's not *that* different to what I'm suggesting.

emersion:
>To bring up a point about terminology, going by what Weston uses, a "head" is the thing which is displaying pixels (e.g. a computer monitor), and an "output" is the logical representation of that. An output can be assigned multiple heads, which can be implemented efficiently by sharing CRTCs where possible.

Right. In my draft proposal about "logical monitors", I've used the terms "head" (as the thing displaying pixels) and "slot" (as the output layout region). In this PR, "head" also describes the thing that displays pixels, but has also been chosen to reduce confusion with `wl_output` and `wlr_output`.

>This type of efficient cloning is something I'd want to see in a protocol like this and in our APIs, but really does require the split concept of a "logical monitor".

I've discussed with Jonas, and that's not really true. The "logical monitor" object is more of a cosmetic thing than a real requirement. You can implement output cloning with the protocol described in this PR: each head with the same mode, position, scale, transform as another head can be cloned.

Cloning is here just a possible optimization, just an implementation detail. In some situations cloning might not be possible at all even if the client requests it.

>All of the modesetting and property setting would happen on the outputs, and be applied to all heads.

It's true that modes are applied to the same CRTC after all.

>we remove the separate mode object and just specify it with width*height@rate

I decided against this to be able to add more features to modes in a later version (e.g. aspect ratio).

>If you want to disable a head, create a new configuration which does not include it.

I've wondered about something like this: should the configuration object always specify the whole configuration?

KMS doesn't have this requirement, so I just decided to only change properties that the user wants to change (diff-like). In an earlier version I had a "unconfigured_head" error which would make it a protocol error to omit a head in a configuration. Maybe it would be better to require the client to be very explicit about the configuration and always specify everything.

I'd prefer not to disable heads that aren't added to the configuration, just to be more explicit.

ascent12:
>only change properties that the user wants to change (diff-like)

This can be a feature of clients by reading in the current configuration and making its changes based off of that. It doesn't need to be part of how the protocol itself works.

>Maybe it would be better to require the client to be very explicit about the configuration and always specify everything.
>I'd prefer not to disable heads that aren't added to the configuration, just to be more explicit.

I'd argue that disabling unmentioned heads is more explicit. It'll remove a bunch of uncertainly and unpredictable behaviour in the DRM backend.
For example, if a new head/output is set up, and I'm unable to configure it without stealing a CRTC from another output, what do I do?
- Do I cause the current request to fail, even though it's the most up-to-date and probably is the output the user wants?
- If I do take it from another output, which one? How do I notify the user of this protocol which one it was, or do they just need to check the configuration again?

emersion:
>It'll remove a bunch of uncertainly and unpredictable behaviour in the DRM backend.

Well, with the current proposal, you send a `disable_head` request to disable a head, so that doesn't change anything about the effective configuration in the end. Or am I missing something?

>do they just need to check the configuration again

This is always the case. The compositor is free to tweak the configuration requested by the client. This is used e.g. for compositors that don't support fractional scaling, and could be used for compositors having other requirements too.

ascent12:
Just to be clear, I'm also talking about changing out proper `wlr_output` at the same time as this protocol.

Basically, knowing the full desired output configuration, it allows me to remove a bunch of annoying edge cases and behaviours in the DRM backend and will probably make the whole thing a lot simpler. None of those "what was active before" and "what is active now" annoyances. I can just try a configuration, use `TEST_ONLY` on it, and tell the user the whole thing worked or "try again", and perhaps even tell them that there were not enough internal resources (CRTCs) to make the request.

emersion:
>This can be a feature of clients by reading in the current configuration and making its changes based off of that. It doesn't need to be part of how the protocol itself works.

Agreed.

emersion:
>Just to be clear, I'm also talking about changing out proper wlr_output at the same time as this protocol.

Yes, to fully take advantage of this protocol it will be necessary to overhaul the backend interface.

I don't plan doing so when implementing this in wlroots though. It'd be way too big. I'll implement it using the current interface first.

But yeah, point is this protocol should be designed to work well with a new, better wlroots API (and with any other compositor, really).

>Basically, knowing the full desired output configuration, it allows me to remove a bunch of annoying edge cases and behaviours in the DRM backend and will probably make the whole thing a lot simpler. None of those "what was active before" and "what is active now" annoyances. I can just try a configuration, use TEST_ONLY on it, and tell the user the whole thing worked or "try again"

Yes, that's exactly the goal.

>and perhaps even tell them that there were not enough internal resources (CRTCs) to make the request.

I'm not too keen on this part. There are a lot of situations where we can't apply the requested configuration, it's not just about a lack of CRTC. For instance, Jonas mentioned that hardware might have some bandwidth limits. Since the KMS interface doesn't allow us to know _why_ a configuration failed, I've just chosen to reflect the same here and make the client test a different one.

ascent12:
>I'm not too keen on this part. There are a lot of situations where we can't apply the requested configuration, it's not just about a lack of CRTC. For instance, Jonas mentioned that hardware might have some bandwidth limits. Since the KMS interface doesn't allow us to know why a configuration failed, I've just chosen to reflect it here and make the client test a different one.

Yeah, I can see that. But in the case of CRTCs, it's not because of the KMS API failing, but because it was never going to work to begin with. A configuration trying to set 4 distinct outputs with only 3 CRTCs available is always going to fail. Maybe we want to tell the client about that, or maybe not.

unstable/wlr-output-management-unstable-v1.xml
```diff
@@ -0,0 +1,444 @@
...
+    SPECIAL, INDIRECT OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN
+    AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION,
+    ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF
+    THIS SOFTWARE.
+  </copyright>
+
+  <description summary="protocol to configure output devices">
+    This protocol exposes interfaces to get and change output device
+    configuration.
```
ddevault: Minor tone nit: "obtain and modify"

unstable/wlr-output-management-unstable-v1.xml
```diff
@@ -0,0 +1,444 @@
...
+        event, so it will become invalid and the client should release any
+        resources associated with it.
+      </description>
+    </event>
+  </interface>
+
+  <interface name="zwlr_output_head_v1" version="1">
+    <description summary="output device">
+      A head is an output device. The difference with wl_output is that heads
+      are advertized even if they are turned off. A head object only advertises
```
ddevault: advertised

unstable/wlr-output-management-unstable-v1.xml
```diff
@@ -0,0 +1,444 @@
...
+        The compositor will destroy the object immediately after sending this
+        event, so it will become invalid and the client should release any
+        resources associated with it.
+      </description>
+    </event>
+  </interface>
+
+  <interface name="zwlr_output_head_v1" version="1">
+    <description summary="output device">
+      A head is an output device. The difference with wl_output is that heads
```
ddevault: The difference between a wl_output and a head is that heads

unstable/wlr-output-management-unstable-v1.xml
```diff
@@ -0,0 +1,444 @@
...
+      are advertized even if they are turned off. A head object only advertises
+      properties and cannot be used directly to change them. In order to update
+      some properties, one needs to create a wlr_output_configuration object.
```
ddevault: In order to update a head's properties

or

In order to update these properties

unstable/wlr-output-management-unstable-v1.xml
```diff
@@ -0,0 +1,444 @@
...
+
+      A head has some read-only properties: mode, name, description and
+      physical_size. These cannot be changed by clients.
+
+      enabled and current_mode are physical properties. Updating them might take
+      some time, depending on hardware limitations.
+
+      position, transform and scale are logical properties. They describe how
+      the output is mapped in the global compositor space.
```
ddevault: I can foresee a future in which transform and scale are physical properties. Best not to make the distinction at all

ddevault: Also, amusing mental picture of the monitors at a desk rearranging themselves on little motorized wheels when a client updates the position

unstable/wlr-output-management-unstable-v1.xml
```diff
@@ -0,0 +1,444 @@
...
+    </description>
+
+    <event name="mode">
+      <description summary="advertise a supported mode">
+        If the head supports modes, this event is sent once per supported mode.
+      </description>
+      <arg name="mode" type="new_id" interface="zwlr_output_mode_v1"/>
```
ddevault: Does this have to create a new resource? It might be better to eliminate the interface, enumerate its properties in this event, and protocol error if the client tries to modeset to an unsupported mode.

unstable/wlr-output-management-unstable-v1.xml
```diff
@@ -0,0 +1,444 @@
...
+      <arg name="name" type="string"/>
+    </event>
+
+    <event name="description">
+      <description summary="head description">
+        This event describes a human-readable description of the head.
+
+        The description is a UTF-8 string with no convention defined for its
+        contents. Examples might include 'Foocorp 11" Display' or 'Virtual X11
+        output via :1'.
```
ddevault: Copypasta the language about not assuming anything from this description

unstable/wlr-output-management-unstable-v1.xml
```diff
@@ -0,0 +1,444 @@
...
+      <arg name="height" type="int" summary="height in millimeters of the output"/>
+    </event>
+
+    <event name="enabled">
+      <description summary="head is enabled or disabled">
+        This event describes whether the head is enabled. A disabled head is not
+        mapped to a region of the global compositor space.
+
+        When a head is disabled, some properties (current_mode, position,
+        transform and scale) are irrelevant.
```
ddevault: Could a compositor do this:

- position
- transform
- current_mode
- enabled (=1)

In that order, where the first three events are meaningful once the final event arrives?

unstable/wlr-output-management-unstable-v1.xml
```diff
@@ -0,0 +1,444 @@
...
+      </description>
+      <arg name="id" type="new_id" interface="zwlr_output_configuration_head_v1"
+        summary="a new object to configure the head"/>
+      <arg name="head" type="object" interface="zwlr_output_head_v1"
+        summary="the head to be enabled"/>
+    </request>
+
+    <request name="disable_head">
+      <description summary="disable a head">
+        Disable a head. The head's properties are irrelevant in this case.
```
ddevault: >The head's properties are irrelevant in this case.

Probably not necessary to mention this as you will have no means with which to set those properties

unstable/wlr-output-management-unstable-v1.xml
emersion: I considered this approach, but this makes modes non-extensible. Future possible features include modes aspect-ratio and interlaced modes, so we need to be able to add new mode properties to the protocol.

unstable/wlr-output-management-unstable-v1.xml
emersion: Yes. All properties in this interface are applied atomically via the `wlr_output_manager.done` event. I'll update the protocol to make this clear.

emersion:
Added custom modes. TBH I'm not sure it's really worth it to have first-class support for outputs with arbitrary modes. Thoughts?

Ongy:
Some output-connector combinations don't report all their modes correctly. So in the long run it might be reasonable to have arbitrary modes on drm to support buggy setups

emersion:
Adding a request to set a custom mode is pretty easy. However properly supporting outputs that don't have modes (e.g. Wayland backend) brings more complexity (see the `arbitrary_modes` event). I wonder if it's worth it, at least in this first version.

emersion:
This is ready for another round of reviews.

unstable/wlr-output-management-unstable-v1.xml
```diff
@@ -0,0 +1,469 @@
...
+      This interface is a manager that allows reading and writing the current
+      output device configuration.
+
+      Output devices that display pixels (e.g. a physical monitor or a virtual
+      output in a window) are represented as heads. Heads cannot be created nor
+      destroyed, but they can be enabled or disabled and their properties can be
+      changed. Each head may have one or more available modes.
+
+      Heads are advertised when the output manager is bound, and whenever they
+      appear.
```
dcz-purism: What about disappearing?

```suggestion
      appear or disappear.
```

emersion: I wanted to mean "new head objects are created and sent" by "heads are advertised". Also heads cannot be removed when the output manager is bound. Maybe we could add "Heads are removed when they disappear"?

unstable/wlr-output-management-unstable-v1.xml
```diff
@@ -0,0 +1,469 @@
...
+
+      Whenever the number of heads or modes changes, the done event will be
+      sent. It carries a serial which can be used in a create_configuration
```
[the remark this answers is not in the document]
emersion: Hmm, not quite. The `done` event is just like `wl_output.done`: it ends a sequence of updates. This means that if the current mode of an output changes, it's not removed and re-advertized. Instead, the `current_mode` event is sent and is followed by a `done` event. Heads are only removed (via their `finished` event) if they disappear (unplugged in the case of a DRM output).

Wording can probably be improved. What do you think?

unstable/wlr-output-management-unstable-v1.xml
```diff
@@ -0,0 +1,469 @@
...
+        configuration changes. However the compositor may emit further events,
+        until the finished event is emitted.
+
+        The client must not send any more requests after this one.
+      </description>
+    </request>
+
+    <event name="finished">
+      <description summary="the compositor has finished with the manager">
+        This event indicates that the compositor is done sending manager events.
```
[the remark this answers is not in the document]
emersion: See the `stop` request.

Maybe we should add a reference to the `stop` request in this description?

unstable/wlr-output-management-unstable-v1.xml
```diff
@@ -0,0 +1,469 @@
...
+      them.
+
+      A head has some read-only properties: mode, name, description and
+      physical_size. These cannot be changed by clients.
+
+      Other properties can be updated via a wlr_output_configuration object.
+
+      Properties sent via this interface are applied atomically via the
+      wlr_output_manager.done event. No guarantees are made regarding the order
+      in which properties are sent.
```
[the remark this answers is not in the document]
emersion: Not all properties are sent with each update. I tried to state this with sentences like "It is only sent if the output is enabled" and "This event is only sent if the head has a physical size".

unstable/wlr-output-management-unstable-v1.xml
```diff
@@ -0,0 +1,469 @@
...
+
+    <event name="name">
+      <description summary="head name">
+        This event describes the head name.
+
+        The naming convention is compositor defined, but limited to alphanumeric
+        characters and dashes (-). Each name is unique among all wlr_output_head
+        objects, but if a wlr_output_head object is destroyed the same name may
+        be reused later. The names will also remain consistent across sessions
+        with the same hardware and software configuration.
```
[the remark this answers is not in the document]
emersion: > must the names change when the configuration changes?

I don't think so.

"Description" will typically contain the make/model/serial, so it could be used instead if you want to detect when the monitor (ie. the physical object) is replaced.

unstable/wlr-output-management-unstable-v1.xml
```diff
@@ -0,0 +1,469 @@
...
+        change over the lifetime of the wlr_output_head object.
+      </description>
+      <arg name="description" type="string"/>
+    </event>
+
+    <event name="physical_size">
+      <description summary="head physical size">
+        This event describes the physical size of the head. This event is only
+        sent if the head has a physical size (e.g. is not a projector or a
+        virtual device).
```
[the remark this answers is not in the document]
emersion: Yeah. Probably better not to make it immutable then.

unstable/wlr-output-management-unstable-v1.xml
```diff
@@ -0,0 +1,469 @@
...
+      </description>
+      <arg name="transform" type="int" enum="wl_output.transform"/>
+    </event>
+
+    <event name="scale">
+      <description summary="current scale">
+        This events describes the scale of the head in the global compositor
+        space. It is only sent if the output is enabled.
+      </description>
+      <arg name="scale" type="fixed"/>
```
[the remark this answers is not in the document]
emersion: >Excuse my ignorance

Eh, you're far from being ignorant :)

>where can I learn the basic types?

I think the docs is the best place: https://wayland.freedesktop.org/docs/html/ch04.html#sect-Protocol-Wire-Format

unstable/wlr-output-management-unstable-v1.xml
```diff
@@ -0,0 +1,469 @@
...
+
+      Properties sent via this interface are applied atomically via the
+      wlr_output_manager.done event. No guarantees are made regarding the order
+      in which properties are sent.
+    </description>
+
+    <event name="size">
+      <description summary="mode size">
+        This event describes the mode size. The size is given in physical
+        hardware units of the output device. This is not necessarily the same as
```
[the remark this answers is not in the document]
emersion: Yes. I've re-used the core protocol (`wl_output`) wording here. I'll ask the exact reason why the term "pixels" or "buffer size" is not used here.

unstable/wlr-output-management-unstable-v1.xml
```diff
@@ -0,0 +1,469 @@
...
+        After this request has been sent, the compositor must respond with an
+        succeeded, failed or cancelled event. Sending a request that isn't the
+        destructor is a protocol error.
+      </description>
+    </request>
+
+    <event name="succeeded">
+      <description summary="configuration changes succeeded">
+        Sent after the compositor has successfully applied the changes or
+        tested them.
```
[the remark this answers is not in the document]
emersion: Even in case of success, a new advertisement of the new state will be triggered. Maybe we should specify it here?

unstable/wlr-output-management-unstable-v1.xml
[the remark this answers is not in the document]
emersion: Turns out there's no real reason this exact wording is used. I still think we should use it for consistency's sake. Though, we could disambiguate it if needed.

```
12:27 <emersion> wl_output question: why is the term "hardware units" used? why isn't "pixel units" or "buffer size" used?
12:27 <emersion> the commit that introduces the change doesn't explain it: https://gitlab.freedesktop.org/wayland/wayland/commit/b6930889044dcd92be5c7499e653c6cb19cddb19
12:28 <emersion> i bet the review is in phabricator ;_;
12:31 <pq> hmm, we never did review in phab... and that commit definitely predates phab
12:32 <pq> as in, I don't think krh was around anymore when Phab was introduced
12:33 <pq> emersion, right, so physical hardware units in this case are indeed pixels as defined by the video mode in use.
12:34 <pq> emersion, notice how the old definition was "in global coordinates".
12:34 <emersion> yeah
12:35 <emersion> thanks
12:35 <pq> emersion, yes, the main point there is that the output size is not in the global coordinate system where surface sizes are
12:35 <pq> or the video mode really, it's talking about video modes, not outputs
```

unstable/wlr-output-management-unstable-v1.xml
dcz-purism: Does that mean that a head is only sent when it changes, or every time any head changes? I'm not sure what "disappear" would mean in the former case.

unstable/wlr-output-management-unstable-v1.xml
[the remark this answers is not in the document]
dcz-purism: While I picked up that the two are related, I wasn't certain that the compositor is the owner of this object.

unstable/wlr-output-management-unstable-v1.xml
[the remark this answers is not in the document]
dcz-purism: I'd go for being too wordy rather than not wordy enough.

unstable/wlr-output-management-unstable-v1.xml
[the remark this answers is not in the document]
dcz-purism: Seeing that my two other notes above were about being confused what is being sent when, I think having a friendly overview somewhere would be useful, e.g. "each time a change happens, events describing this change must be sent, followed by the done event", or "a disappearing head causes the following to happen", or "if an event does not appear between done invokations, then corresponding values stay unchanged". In my ideal world, that would be supported by relevant "must" and "may"s next to the events, although I like being wordy.

unstable/wlr-output-management-unstable-v1.xml
emersion: Okay, so let's go through all possible situations.

### A head "appears"

A new output is plugged in:

```
zwlr_output_manager_v1.head(new id @42)
zwlr_output_head_v1@42.name("VGA-1")
zwlr_output_head_v1@42.description("A new output i just plugged in")
zwlr_output_head_v1@42.enabled(1)
…
zwlr_output_manager_v1.done(64)
```

### A head changes

For instance a head gets disabled. In this case no head "appears" or "disappears". Thus no head object is created or destroyed.

```
zwlr_output_head_v1@42.enabled(0)
zwlr_output_manager_v1.done(65)
```

### A head "disappears"

The output is unplugged.

```
zwlr_output_head_v1@42.finished()
zwlr_output_manager_v1.done(66)
```

### Or any combination of the above

When multiple changes happen at the same time, they are all grouped before a single `done` event. For instance undocking a laptop will make many outputs go away at the same time.

unstable/wlr-output-management-unstable-v1.xml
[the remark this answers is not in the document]
emersion: The concept of "owner" is difficult to define. Some interfaces expose a "destroy" request that ask the compositor to destroy the object. Other interfaces have a "finished" event that notify the client that the object doesn't exist anymore. In both cases, it's not very clear who is the "owner". It's always a matter of making sure both sides agree an object is still alive or has been destroyed.

unstable/wlr-output-management-unstable-v1.xml
dcz-purism: Thanks, this makes sense, and indeed you put notes in head::finished as well as head::enabled. I'm going to be fine with this if there's a note that mentions that not all state needs to be resent at every change.

emersion:
Thanks for your comments @dcz-purism! I've made changes to address your comments, let me know what you think and if something needs to be disambiguated.

ddevault:
Nice, I'll have another look, too.

ddevault:
Thanks!

PR #43, opened 2019-04-12
Clearify exclusive zone

wmww:
Describe under what anchors the exclusive zone has an effect.

### DIFF
--- a/unstable/wlr-layer-shell-unstable-v1.xml
+++ b/unstable/wlr-layer-shell-unstable-v1.xml
@@ -138,8 +138,8 @@
         to only two parallel edges or anchored to all edges, a positive value
         will be treated the same as zero.
 
-        A positive zone is the number of surface-local coordinates to consider
-        exclusive.
+        A positive zone is the distance from the edge in surface-local
+        coordinates to consider exclusive.
 
         Surfaces that do not wish to have an exclusive zone may instead specify
         how they should interact with surfaces that do. If set to zero, the
--- a/unstable/wlr-layer-shell-unstable-v1.xml
+++ b/unstable/wlr-layer-shell-unstable-v1.xml
@@ -127,14 +127,19 @@
 
     <request name="set_exclusive_zone">
       <description summary="configures the exclusive geometry of this surface">
-        Requests that the compositor avoids occluding an area of the surface
-        with other surfaces. The compositor's use of this information is
+        Requests that the compositor avoids occluding an area with other
+        surfaces. The compositor's use of this information is
         implementation-dependent - do not assume that this region will not
         actually be occluded.
 
-        A positive value is only meaningful if the surface is anchored to an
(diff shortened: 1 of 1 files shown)


unstable/wlr-layer-shell-unstable-v1.xml
```diff
@@ -127,14 +127,19 @@
 
     <request name="set_exclusive_zone">
       <description summary="configures the exclusive geometry of this surface">
-        Requests that the compositor avoids occluding an area of the surface
-        with other surfaces. The compositor's use of this information is
```
wmww: This could be interpreted as meaning an exclusive zone larger than the surface should be ignored. If this is intentional, it should be explicitly specified. I'm assuming that is not the meaning.

unstable/wlr-layer-shell-unstable-v1.xml
```diff
@@ -127,14 +127,19 @@
...
         actually be occluded.
 
-        A positive value is only meaningful if the surface is anchored to an
-        edge, rather than a corner. The zone is the number of surface-local
-        coordinates from the edge that is considered exclusive.
+        A positive value is only meaningful if the surface is anchored to one
+        edge or an edge and both perpendicular edges. If the surface is not
+        anchored, anchored to only two perpendicular edges (a corner), anchored
+        to only two parallel edges or anchored to all edges, a positive value
+        will be treated the same as zero.
```
emersion: Maybe this should be a protocol error

emersion: (Would require a version bump, and can be done separately)

unstable/wlr-layer-shell-unstable-v1.xml
```diff
@@ -127,14 +127,19 @@
...
+
+        A positive zone is the number of surface-local coordinates to consider
```
emersion: Alternative wording: "distance from the edge in surface-local coordinates"

wmww: Much better

unstable/wlr-layer-shell-unstable-v1.xml
wmww: I don't have an opinion on this

ddevault: A protocol error would be nice for a future patch, agreed.

ddevault:
Thanks!

unstable/wlr-layer-shell-unstable-v1.xml
wmww: Does that mean it will be required to set the exclusive zone to zero in the same commit as you change the anchor to one that does not allow an exclusive zone?

emersion: Yeah, that would make sense

Issue #45, opened 2019-04-25
layer-shell: Notify of fullscreen/maximize


AdrianVovk:
Hello,

Currently, layer-shell has no signal to indicate that some surface has gone full-screen and will cover the panel. For simple panels, this is fine. I, however, am trying to build a DE with Wayfire, and it would be great to use layer-shell (if my users want to run my DE's components on Sway for tiling, for example). I have functionality where the super key will open the application menu, and the panel pops up temporarily over fullscreened surfaces. I also have a Super + T binding that pops up the panel for a few hundred ms to check the time or whatnot. layer-shell does not allow me in any way to know when a surface has gone fullscreen, so I cannot do any of this. If this is implemented separately from layer-shell, there needs to be some way for clients to tell the compositor that they handle fullscreen windows and shouldn't be forced below them.

It also seems fitting that layer-shell could notify the panel when a client is maximized. My panel is translucent when nothing is tiled/maximized, but should go fully opaque when something is. I'm certain this is a common usecase (Pantheon, GNOME, etc). This could go into a separate protocol if need be.

Thank you!

emersion:
Window management is out-of-scope for layer-shell.

However another protocol already exposes maximized windows: https://github.com/swaywm/wlr-protocols/blob/master/unstable/wlr-foreign-toplevel-management-unstable-v1.xml

@ammen99 Maybe fullscreen could be added too?

ammen99:
@emersion The behavior that we want here is something else, i.e "If the current workspace has a fullscreen window, hide the panel". It depends on the "focused" workspace and on the view state.

emersion:
Would foreign-toplevel-management + a workspace protocol be enough?

ammen99:
Probably yes, with an additional fullscreen mode for foreign-toplevel (I'll send a patch for that soon)

AdrianVovk:
How would I signal compositors to not place my panel under fullscreened surfaces?

ammen99:
@AdrianVovk by placing it in the overlay layer?

ammen99:
Actually I think we don't even need the workspace protocol for this. The panel can just check whether the currently activated toplevel is fullscreen, and if yes, then autohide.

However if the panel also needs to know if there is a maximized window on the current workspace, it will need a workspace protocol as well.

emersion:
If you have two screens and a non-activated fullscreen toplevel on one of them, will that still work?

ammen99:
[...]

PR #46, opened 2019-04-26
foreign-toplevel-management: add support for fullscreening toplevels

ammen99:
This can be used for fullscreening an app from a taskbar, and potentially also for panels to autohide themselves in case a fullscreen toplevel appears. Related to #45

### DIFF
--- a/unstable/wlr-foreign-toplevel-management-unstable-v1.xml
+++ b/unstable/wlr-foreign-toplevel-management-unstable-v1.xml
@@ -25,7 +25,7 @@
     THIS SOFTWARE.
   </copyright>
 
-  <interface name="zwlr_foreign_toplevel_manager_v1" version="1">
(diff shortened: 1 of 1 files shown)


AdrianVovk:
@ammen99 This doesn't seem solve emersion's issue. How will we handle that?

We need a state that basically says "I'm on top of everything else but not necessarily active", or "I was active but now the output is inactive" or something like that. Maybe just giving the z-index would be good enough? It would be nice to have another state, though

emersion:
>We need a state that basically says "I'm on top of everything else but not necessarily active", or "I was active but now the output is inactive" or something like that. Maybe just giving the z-index would be good enough? It would be nice to have another state, though

I think this is a different issue and it should be addressed separately. I'm still not sure what's the best way to handle it.

AdrianVovk:
Well in that case, LGTM

emersion:
Thanks!

PR #51, opened 2019-07-21
layer-shell: Fix conflict with C++

plfiorini:
namespace is a C++ keyword and this makes this protocol incompatible,
rename the argument in a way that doesn't conflict.


ddevault:
No thanks. You should use a scanner designed for C++.

plfiorini:
Or write one :)
With QtWayland we already have a C++ generator but it relies on C code generated by `wayland-scanner` and that's the code that is causing build issues.
For now I have renamed `namespace` in my local copy.

PR #52, opened 2019-07-28
foreign-toplevel-management: Report the surface's parent

dos1:
Some window switchers, like mobile ones, may want to skip or group dialogs and other child windows on their window lists. Right now the only way to group toplevels is to use app_id, which will end up grouping too much in case of multiple toplevels with separate documents. Reporting on the toplevel's parent makes it easy to distinguish separate window stacks and act accordingly.

### DIFF
--- a/unstable/wlr-foreign-toplevel-management-unstable-v1.xml
+++ b/unstable/wlr-foreign-toplevel-management-unstable-v1.xml
@@ -25,7 +25,7 @@
     THIS SOFTWARE.
   </copyright>
 
-  <interface name="zwlr_foreign_toplevel_manager_v1" version="2">
+  <interface name="zwlr_foreign_toplevel_manager_v1" version="3">
     <description summary="list and control opened apps">
       The purpose of this protocol is to enable the creation of taskbars
       and docks by providing them with a list of opened applications and
@@ -68,7 +68,7 @@
     </event>
   </interface>
 
-  <interface name="zwlr_foreign_toplevel_handle_v1" version="2">
+  <interface name="zwlr_foreign_toplevel_handle_v1" version="3">
     <description summary="an opened toplevel">
       A zwlr_foreign_toplevel_handle_v1 object represents an opened toplevel
       window. Each app may have multiple opened toplevels.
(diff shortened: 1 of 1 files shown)


ddevault:
Note that this will only work if the client has bound an object to the parent topleve.

emersion:
>Note that this will only work if the client has bound an object to the parent toplevel.

Since toplevels are not globals, but are server-created objects via the `toplevel` event on the manager, it should be fine.

However the docs should specify what happens if the client has destroyed the toplevel that is the parent. No event? Event with null?

I'd also like to see a working implementation, just to make sure there are no issues we missed.

AdrianVovk:
Ping? This would be really great to have in the proto

ddevault:
Send patch?

dkondor:
Hi,

I'm interested in having this functionality and started working on an implementation.
I have a working example with wlroots + wayfire + cairo-dock, as a combination of these three branches:
https://github.com/dkondor/wlroots/tree/foreign-toplevel-report-parent
https://github.com/dkondor/wayfire/tree/foreign-toplevel-report-parent
https://github.com/dkondor/cairo-dock-core/tree/wayland_egl

So far, what I've tested is the simple rule that child views should not be displayed in the taskbar as separate icons. I have not tested more complicated functionality (grouping / mapping views together). I'm happy to create other examples beside cairo-dock which is a complex case, if needed. Should I create separate PRs for wlroots and wayfire?

Thanks!

ammen99:
> Hi,
> 
> I'm interested in having this functionality and started working on an implementation.
> I have a working example with wlroots + wayfire + cairo-dock, as a combination of these three branches:
> https://github.com/dkondor/wlroots/tree/foreign-toplevel-report-parent
> https://github.com/dkondor/wayfire/tree/foreign-toplevel-report-parent
> https://github.com/dkondor/cairo-dock-core/tree/wayland_egl
> 
> So far, what I've tested is the simple rule that child views should not be displayed in the taskbar as separate icons. I have not tested more complicated functionality (grouping / mapping views together). I'm happy to create other examples beside cairo-dock which is a complex case, if needed. Should I create separate PRs for wlroots and wayfire?
> 
> Thanks!

Yes, the protocol is waiting for implementation :)

ammen99:
> However the docs should specify what happens if the client has destroyed the toplevel that is the parent. No event? Event with null?

IMO no event makes more sense than event with null. After all, null events means no parent. Otherwise LGTM.

emersion:
OK, no event makes sense to me. After all, if the client has destroyed the toplevel, it's on them.

Can you add it to the protocol description?

dos1:
Is this working okay?

emersion (commented):
Yup, protocol changes LGTM. Waiting on the wlroots pull request to be reviewed before merging this.

emersion (approved):

emersion:
Thanks!

PR #53, opened 2019-08-05
Add a protocol for monitoring damage regions

any1:
This is very useful for VNC.


ddevault:
I think I would rather see this as an extension to screencopy, so you can more directly correlate damage with frames.

any1:
Screencopy only copies a frame upon request. What do you do with damage between requests? Accumulate it?
Or do we maybe extend screencopy so that it can stream buffers with damage?

ddevault:
I'm not sure of the specifics in how screencopy is affected. Do you think you can experiment it with a bit and draft something which might work?

any1:
Sure. How does it sound to accumulate damage per client and have it all sent along with the next request? The accumulator would then be zeroed after the request has been serviced.

PR #54, opened 2019-08-09
screencopy: Add request and event for coarse damage reporting

any1:


ammen99:
Can't we pass a `wl_region` for the damage?

any1:
How do you read the contents of a wl_region? There does not seem to be a public interface for getting the coordinates.

ammen99:
Oh, you're right. It's not a good idea then.

ddevault:
This change is not backwards compatible, and it ought to be.

any1:
I thought that "unstable" meant that we didn't care. I'll try to figure out a way to make it backwards compatible.

any1:
The interface is now backwards compatible, but a client using the new features has no way of knowing if the server supports them. Is that what the version attribute is for?

ammen99:
@any1 Yes, version is used exactly for this purpose.
 
Aside from that, I think the general convention is to add the new events/requests at the end of the interface (i.e after the older events and requests)

unstable/wlr-screencopy-unstable-v1.xml
```diff
@@ -120,6 +120,13 @@
       <arg name="buffer" type="object" interface="wl_buffer"/>
     </request>
 
+    <request name="copy_with_damage">
```
ammen99: Needs a `since = ` attribute.

unstable/wlr-screencopy-unstable-v1.xml
```diff
@@ -170,6 +185,21 @@
       </description>
     </event>
 
+    <event name="damage">
```
ammen99: Needs a `since = ` attribute.

any1:
Yeah, the request should have gone below the "destroy" request for the server interface to be compatible. I've moved the additions to the bottom now and added the attributes that you requested.

unstable/wlr-screencopy-unstable-v1.xml
```diff
@@ -160,6 +160,14 @@
...
       <arg name="tv_nsec" type="uint"
            summary="nanoseconds part of the timestamp"/>
+      <arg name="damage_x" type="uint"
+           summary="current x coordinates for damage"/>
+      <arg name="damage_y" type="uint"
+           summary="current y coordinates for damage"/>
+      <arg name="damage_width" type="uint"
+           summary="current width of damage"/>
+      <arg name="damage_height" type="uint"
+           summary="current height of damage"/>
```
ddevault: Adding arguments to an existing event is not backwards-compatible. Add a new damage event instead.

any1: I forgot to remove it when I added the damage event. It is removed now.

ddevault:
Can we see an implementation of the updated protocol somewhere?

any1:
Nope, I wanted to have this decided before I started implementing. Are these interface changes good enough as is? I've already implemented the damage-stream interface that I proposed before, so I have an idea about how to implement it. See: https://github.com/any1/wlroots/blob/damage-stream/types/wlr_damage_stream_v1.c

ddevault:
>Nope, I wanted to have this decided before I started implementing

That's not how this works. We use implementations to prove that protocol designs are correct.

any1:
https://github.com/any1/wlroots/commit/1a28d0e5c20b6c3ea635aa286924d3c973fe66c6

ddevault:
And a client?

emersion:
Can you open a PR for the wlroots patch?

any1:
There is a client implemented here: https://github.com/any1/wvnc/blob/screencopy-damage/main.c

mstoeckl:
It would probably be better to expose more fine grained damage tracking in this interface, because providing only the extents of the damaged region can significantly overestimate the covered area. For example, if there is a status bar change in the bottom right of the screen, and a new text character in the top left, then the damage extent rectangle will cover almost the entire frame, even when only a few pixels have changed.

(On the wlroots side of things, using `pixman_region32_rectangles` will produce a set of [disjoint](https://gitlab.freedesktop.org/pixman/pixman/blob/master/pixman/pixman-region.c#L122) rectangles that are suitable to be passed as damage.)

Wayland has a rarely-used `array` type which would be useful to pass the damaged region in a single message; the only downside is the message size limit, which would in practice limit damage specifications to <255 rectangles.

any1:
I decided to use the extents only because it makes things a lot simpler while still offering significant improvements. The most useful aspect of this is actually to reduce screen copying while there is little happening on the screen. Without damage notification, the user has to continually poll and this causes wasted CPU cycles.
I will add rectangles if there is consensus for it, but I'm not sure if the increase in complexity pays off.

any1:
This PR seems to have stalled a bit. What can I do to help move the issue forward?

ddevault:
You just have to be patient. You haven't been forgotten about, we're just busy people and most of us work on this in our spare time.

ddevault:
Regarding fine-grained damage, for now I would prefer to keep this simple.

ddevault:
Tested this and it seems to work fine. I'm +1 on this protocol change assuming the wlr work is cleaned up.

emersion:
I'm not sure about merging this. Once this is merged, we can't go back, meaning future versions won't be able to cleanly implement fine-grained damage on top of this.

I'd like to think about only copying the damaged regions, just to understand how we can make the protocol future-proof.

ddevault:
Well, this is an unstable protocol, so we could add another event later and consolidate when it becomes stable. I would like to look into only copying the affected parts of the screen, though, that's a good idea.

unstable/wlr-screencopy-unstable-v1.xml
```diff
@@ -175,5 +175,27 @@
...
+    <event name="damage" since="2">
+      <description summary="carries the coordinates of the damaged region">
+        This event is sent right before the ready event when copy_with_damage is
+        requested.
+
+        The arguments describe a box around the area that has changed since the
+        last copy request that was derived from the current screencopy manager
+        instance.
+      </description>
+      <arg name="x" type="uint" summary="current x coordinates"/>
```
emersion: "current" is a little vague. Maybe replace with "damaged"?

unstable/wlr-screencopy-unstable-v1.xml
```diff
@@ -175,5 +175,27 @@
...
+      <description summary="copy the frame when it's damaged">
+        Same as copy, except it waits until there is damage to copy.
+      </description>
+      <arg name="buffer" type="object" interface="wl_buffer"/>
+    </request>
+
+    <event name="damage" since="2">
+      <description summary="carries the coordinates of the damaged region">
+        This event is sent right before the ready event when copy_with_damage is
+        requested.
```
emersion: This doesn't specify this is only sent once. Either we should make it mandatory to send it only once, either we should allow sending it multiple times, describing the fine-grained region.

any1: Yes, we could add fine-grained damage later by specifying that this may be called multiple times.

any1:
> You just have to be patient. You haven't been forgotten about, we're just busy people and most of us work on this in our spare time.

Sorry. I didn't mean to be pushy. Because I know that you are busy people, I just wanted to know if I could do something to make this easier. For now, I will make the required changes. Thanks for the review!

any1:
I've added a stipulation that copy_with_damage is only required to copy the damaged region. Does that make this interface future-proof in this regard?

To be clear: The damaged region will be copied to the appropriate region within the buffer, not to the start.

unstable/wlr-screencopy-unstable-v1.xml
```diff
@@ -79,7 +80,7 @@
     </request>
   </interface>
 
-  <interface name="zwlr_screencopy_frame_v1" version="1">
+  <interface name="zwlr_screencopy_frame_v1" version="2">
```
emersion: The version is inherited from the object that created the child object. Thus all interface versions in the protocol needs to be bumped.

emersion:
Make it possible to send the damage event multiple times, for a fine-grained region: if possible, I'd prefer to do this in the current protocol version. It is a simple change, the compositor can send too much damage. It just reduces complexity regarding versions: when each version has a different behavior for the same events/requests it becomes quite hard to follow the flow.

Plans for the future: the compositor can copy only the damaged region, and the client can ask that at least a given region will be copied over.

any1:
@emersion and I discussed partial copying on IRC and agreed that we will add the required interface changes later.

The main issue with partial copying is that multi-buffering on the client side becomes tricky:
> 13:02 < emersion> if you copy from the buffer sent to the compositor, it's racy, the compositor is modifying it
> 13:02 < emersion> the buffer being processed is the one that needs previous damage to be copied over
> 13:03 < emersion> so you need a copy of the last buffer, to be able to fill the gaps in the new one
> 13:04 < emersion> alternatively, the client could indicate which regions it wants the compositor to copy over
> 13:05 < emersion> "copy at least this region, plus the damage"

unstable/wlr-screencopy-unstable-v1.xml
any1: Hmm, actually, I changed it in the implementation so that the version comes from the interface struct. Doesn't that make more sense?

emersion: No, because then the object will always have the same version, regardless of what version is specified by the client.

emersion:
The only issue I'm seeing with this PR is that it's racy. If the client takes too much time sending requests, it may miss some damage. This isn't a big deal with the old `copy` request (the client will just drop frames it can't process), however this may result in a corrupted video stream with this new `copy_with_damage` request.

any1:
The race is handled by accumulating damage between requests from the same manager instance.

ddevault:
Maybe we should add a damage serial, and a request which flushes (acknowledges) damage up through that serial?

any1:
Sorry, I'm going to need further explanation for the "damage serial" idea.

ddevault:
Damage doesn't stop accumulating until the client acknolwedges it, then it's cleared up to the point where they acknowledged it. So each damage event contains all of the damage since the last time the client acked it. The serial is used to identify how much damage the client has processed.

ddevault:
Nevermind, this isn't necessary. I see what you're getting at with accumulating on the client.

any1:
Currently, the accumulated damage is flushed when the client requests a new frame and in doing so it implicitly acknowledges that it has processed the previous frame and its damage. What is the benefit of making the acknowledgement explicit?

unstable/wlr-screencopy-unstable-v1.xml
```diff
@@ -175,5 +176,31 @@
         Destroys the frame. This request can be sent at any time by the client.
       </description>
     </request>
+
+    <request name="copy_with_damage" since="2">
```
emersion: Actually, can you add a `<!-- Version 2 additions -->` comment above this request?

any1:
> Can you turn your wlroots changes into a pull request before this is merged?

They are not a PR already?

ddevault:
Ah, the pull request was further down in my review queue. Apologies.

ddevault:
Thanks!

Issue #36, opened 2019-03-01
Virtual pointer protocol


emersion:
We have a virtual keyboard protocol, but we're missing a virtual pointer one. Could be useful for remote control and remote desktop.

emersion:
Protocol draft by @atx: https://github.com/atx/wlr-protocols/blob/master/unstable/wlr-virtual-pointer-unstable-v1.xml

ddevault:
I'm not sure why virtual pointers would need anything beyond what, for example, the wayland backend provides.

ddevault:
Hm, I'm not fond of that idea. The Wayland protocol does not already encode libinput, we would be making it less portable.

swick:
It does not add a dependency on libinput but evdev (yes, the name is misleading in that way).

pq mentioned that this would not work for use cases where you want to move the pointer to a specific location since the events (probably) get modified in some form by the compositor. It seems like both solution have something that the other can't provide.

ddevault:
Right, it can't just be the inverse of wl_pointer. I think we should aim for the inverse of wl_pointer plus an absolute motion event modelled after the libinput API.

ddevault:
Perhaps plus a suggestion of which wl_output the pointer prefers to be on? This is something we can theoretically get out of udev for real input devices but I don't think we have the code.

emersion:
>Perhaps plus a suggestion of which wl_output the pointer prefers to be on?

Maybe. The output would be useful for absolute events only.

Note that pointers are generally not mapped to outputs. Touch devices and tablets are.

> This is something we can theoretically get out of udev for real input devices but I don't think we have the code.

No, there's no way to get this from udev. Other compositors resort to heuristics (e.g. trying to map a touch device to the first integrated display), but that doesn't work in all cases.

ddevault:
>Note that pointers are generally not mapped to outputs. Touch devices and tablets are.

This is only true of the traditional desktop paradigm, but for many use-cases (wayland backend, x11 backend, vnc) this is not true.

emersion:
Implemented in https://github.com/swaywm/wlr-protocols/pull/58

PR #58, opened 2019-09-25
Add wlr-virtual-pointer-unstable-v1

atx:
See #36 for previous discussion

### DIFF
--- a/unstable/wlr-virtual-pointer-unstable-v1.xml
+++ b/unstable/wlr-virtual-pointer-unstable-v1.xml
@@ -1,0 +1,128 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<protocol name="wlr_virtual_pointer_unstable_v1">
+  <copyright>
+    Copyright © 2019 Josef Gajdusek
+
+    Permission is hereby granted, free of charge, to any person obtaining a
+    copy of this software and associated documentation files (the "Software"),
+    to deal in the Software without restriction, including without limitation
+    the rights to use, copy, modify, merge, publish, distribute, sublicense,
+    and/or sell copies of the Software, and to permit persons to whom the
+    Software is furnished to do so, subject to the following conditions:
+
+    The above copyright notice and this permission notice (including the next
+    paragraph) shall be included in all copies or substantial portions of the
+    Software.
+
+    THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+    IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+    FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
+    THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+    LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+    FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+    DEALINGS IN THE SOFTWARE.
+  </copyright>
+
+  <interface name="zwlr_virtual_pointer_v1" version="1">
+    <description summary="virtual pointer">
+      This protocol allows clients to emulate a physical pointer device. The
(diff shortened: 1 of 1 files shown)


ammen99:
I think floating point numbers should be used for most events, just as they are used in the core wayland protocol.

atx:
> I think floating point numbers should be used for most events, just as they are used in the core wayland protocol.

The wayland protocol does not use floats though? It uses 32 bit fixed point numbers (wl_fixed_t). This is problematic, as we do not want to lose precision (scaling from 0.0 to 1.0 would leave us with just 8 bits for example).

There is also no WL_FIXED_MAX constant, so there is no "nice" way of telling clients to just use the full available precision.

emersion:
For the relative motion events, I think a 24.8 fixed point value in surface-local coords is fine, as [whot said](https://github.com/swaywm/wlr-protocols/issues/36#issuecomment-534521576).

Naturally for absolute motion events things are different.

ammen99:
I saw the mention of surface-local coordinates, but there are no surfaces used in this protocol. Maybe use another description?

atx:
Good point - the compositor could scale/transform the currently-focused surface in some way. We want to move the cursor in output coordinate space instead.

unstable/wlr-virtual-pointer-unstable-v1.xml
```diff
@@ -0,0 +1,124 @@
...
+    <description summary="virtual pointer">
+      This protocol allows clients to emulate a physical pointer device. The
+      requests are mostly mirror opposites of those specified in wl_pointer.
+    </description>
+
+    <request name="motion">
+      <description summary="pointer relative motion event">
+        The pointer has moved by a relative amount to the previous request.
+
+        Values are in output coordinate space.
```
ammen99: Sorry, maybe I am dumb, but no output has been defined in the protocol as well. Do you mean the output the cursor is on? Or something else?

atx: Right, changed that to "global compositor space", which is a term used by the core wayland protocol.

unstable/wlr-virtual-pointer-unstable-v1.xml
```diff
@@ -0,0 +1,124 @@
...
+
+    <request name="motion_absolute">
+      <description summary="pointer absolute motion event">
+        The pointer has moved in an absolute coordinate frame.
+
+        Value can range from 0 to UINT_MAX
+      </description>
+      <arg name="time" type="uint" summary="timestamp with millisecond granularity"/>
+      <arg name="x" type="uint" summary="position on the x-axis"/>
+      <arg name="y" type="uint" summary="position on the y-axis"/>
```
ddevault: We should probably do this on a scale from 0..1, with fixed

ddevault:
LGTM aside from the one issue

Thanks :smile:

unstable/wlr-virtual-pointer-unstable-v1.xml
atx: I might be misunderstanding something, but that would leave us with only 8 bits of precision wouldn't it? That's nowhere enough for absolute pointer device.

ddevault: Yeah, that wouldn't work. Good point. But we still don't have a good way to specify the range here. Consider that for the use-case of VNC we can describe it in pixels, but for the use-case of e.g. a drawing tablet we can only describe it in the coordinate space of the device. Perhaps we should add an x_extent and y_extent arg to this to define the coordinate space of the event?

atx: Okay, that seems reasonable.

My original idea was to specify the extents while creating the pointer object, which I wanted to avoid as it introduces unnecessary state.

ddevault:
Needs a client and server implementation now.

atx:
Most of the plumbing is done in my [sway](https://github.com/atx/sway/tree/virtual-pointer)/[wlroots](https://github.com/atx/wlroots/tree/virtual-pointer) repos. I am going to submit a PR later this week.

ddevault:
Excellent, I'm looking forward to it.

ddevault:
Updated to make wl_seat a suggestion on pointer create.

ddevault:
Future direction: it would be helpful to be able to suggest an output the pointer is associated with, which the compositor can use as a hint for interpreting absolute events.

emersion:
Thanks!

PR #59, opened 2019-10-04
layer_shell: set layer of existing surface

Timidger:


unstable/wlr-layer-shell-unstable-v1.xml
```diff
@@ -236,6 +236,13 @@
       </description>
     </request>
 
+    <request name="set_layer" since="2">
+      <description summary="change the layer of the surface">
+        Change the layer that the surface is rendered on.
```
emersion: This should specify that this is double-buffered, see the other requests.

I'd also be more comfortable with appending this at the end of the interface, with a `<!-- Version 2 additions -->` comment.

Timidger: Done, also updated [these lines](https://github.com/swaywm/wlr-protocols/pull/59/files#diff-8d1419da79a9ab9a0cf9ab943a8d1b8fR93-R95) to include `layer` as something that is double buffered.

ddevault:
LGTM, @emersion?

Issue #56, opened 2019-08-27
Add protocol to toggle output dpms


agx:
Since #38 doesn't deal with output DPMS it would be good to be able to toggle an output's power save mode via a wayland protocol. I have put together a prototype in phoc/phosh to see how things go:

https://source.puri.sm/Librem5/phoc/merge_requests/60/diffs#808a606abe792c793cb5ed64b63f99d6dcb86a60_77_82

If it makes sense I'd like to turn it into a wlr-output-dpms so phosh can continue to run with other compositors and other shell's/daemons don't need to grow custom protocol. In this case  should we still call it DPMS? Modern drm mostly treats it as legacy and does on/off only. However keeping different states gives the compositor more room like e.g. only turning off the backlight on `standby` and fully turning off the display pipeline on `off`. This is something e.g. needed on a phone since the time to bring back the display pipeline from off is too long when the user expects a quick unblank.

ddevault:
Do we need a whole protocol for this or can't we just update wlr-output-management?

emersion:
>Do we need a whole protocol for this or can't we just update wlr-output-management?

-1. Clients interested in output management (deamons listening to hotplugs and output configuration apps) don't intersect with clients interested in power management (e.g. swayidle). Moreover output management deals with disabled outputs, power management doesn't.

> However keeping different states gives the compositor more room like e.g. only turning off the backlight on standby and fully turning off the display pipeline on off. This is something e.g. needed on a phone since the time to bring back the display pipeline from off is too long when the user expects a quick unblank.

What would the logic be in the client to decide when to transition from suspend to off? Couldn't it be implemented in the compositor?

agx:
> What would the logic be in the client to decide when to transition from suspend to off? Couldn't it be implemented in the compositor?

I was thinking to keep policy out of the compositor (go from standby to off after n minutes, standby is enough when not running on battery, suspend faster when only background tasks are running). While all of this could be done in the compositor other systems have more knowledge about this.

emersion:
>I was thinking to keep policy out of the compositor

This makes sense, but I'm wondering whether it's fine to have suspend in a general protocol. What kind of API would the compositor have to shut down the whole display pipeline? Is this somewhat exposed via DRM?

ascent12:
I'm not really convinced about the general applicability of this. The option of having various states (e.g. affecting backlight control) is not something that applies to desktops, and as you mentioned, they are completely useless in modern DRM, which reflects how LCDs work.

[...]

PR #61, opened 2019-10-25
Add output power management protocol

agx:
This specifies how to control power management modes of outputs.

Closes: #56

### DIFF
--- a/unstable/wlr-output-power-management-unstable-v1.xml
+++ b/unstable/wlr-output-power-management-unstable-v1.xml
@@ -1,0 +1,128 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<protocol name="wlr_output_power_management_unstable_v1">
+  <copyright>
+    Copyright © 2019 Purism SPC
+
+    Permission is hereby granted, free of charge, to any person obtaining a
+    copy of this software and associated documentation files (the "Software"),
+    to deal in the Software without restriction, including without limitation
+    the rights to use, copy, modify, merge, publish, distribute, sublicense,
+    and/or sell copies of the Software, and to permit persons to whom the
+    Software is furnished to do so, subject to the following conditions:
+
+    The above copyright notice and this permission notice (including the next
+    paragraph) shall be included in all copies or substantial portions of the
+    Software.
+
+    THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+    IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+    FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
+    THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+    LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+    FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+    DEALINGS IN THE SOFTWARE.
+  </copyright>
+
+  <description summary="Control power management modes of outputs">
+    This protocol allows clients to control power management modes
+    of outputs that are currently part of the compositor space. The
+    intent is to allow special clients like desktop shells to power
+    down outputs when the system is idle.
+
+    To modify outputs not currently part of the compositor space see
+    wlr-output-management.
+
+    Warning! The protocol described in this file is experimental and
+    backward incompatible changes may be made. Backward compatible changes
+    may be added together with the corresponding interface version bump.
+    Backward incompatible changes are done by bumping the version number in
+    the protocol and interface names and resetting the interface version.
+    Once the protocol is to be declared stable, the 'z' prefix and the
+    version number in the protocol and interface names are removed and the
+    interface version number is reset.
+  </description>
+
+  <interface name="zwlr_output_power_manager_v1" version="1">
+    <description summary="manager to create per-output power management">
+      This interface is a manager that allows creating per-output power
+      management mode controls.
+    </description>
+
+    <request name="get_output_power">
+      <description summary="get a power management for an output">
+        Create a output power management mode control that can be used to
+        adjust the power management mode for a given output.
+      </description>
+      <arg name="id" type="new_id" interface="zwlr_output_power_v1"/>
+      <arg name="output" type="object" interface="wl_output"/>
+    </request>
+
+    <request name="destroy" type="destructor">
+      <description summary="destroy the manager">
+        All objects created by the manager will still remain valid, until their
+        appropriate destroy request has been called.
+      </description>
+    </request>
+  </interface>
+
+  <interface name="zwlr_output_power_v1" version="1">
+    <description summary="adjust power management mode for an output">
+      This object offers requests to set the power management mode of
+      an output.
+    </description>
+
+    <enum name="mode">
+      <entry name="off" value="0"
+             summary="Output is turned off."/>
+      <entry name="on" value="1"
+             summary="Output is turned on, no power saving"/>
+    </enum>
+
+    <enum name="error">
+      <entry name="invalid_mode" value="1" summary="inexistent power save mode"/>
+    </enum>
+
+    <request name="set_mode">
+      <description summary="Set an outputs power save mode">
+        Set an output's power save mode to the given mode. The mode change
+        is effective immediately. If the output does not support the given
+        mode a failed event is sent.
+      </description>
+      <arg name="mode" type="uint" enum="mode" summary="the power save mode to set"/>
+    </request>
+
+    <event name="mode">
+      <description summary="Report a power management mode change">
+        Report the power management mode change of an output.
+
+        The mode event is sent after an output changed its power
+        management mode. The reason can be a client using set_mode or the
+        compositor deciding to change an output's mode.
+        This event is also sent immediately when the object is created
+        so the client is informed about the current power management mode.
+      </description>
+      <arg name="mode" type="uint" enum="mode"
+           summary="the output's new power management mode"/>
+    </event>
+
+    <event name="failed">
+      <description summary="object no longer valid">
+        This event indicates that the output power management mode control
+        is no longer valid. This can happen for a number of reasons,
+        including:
+        - The output doesn't support power management
+        - Another client already has exclusive power management mode control
+          for this output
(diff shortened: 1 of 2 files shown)


unstable/wlr-output-power-management-unstable-v1.xml
```diff
@@ -0,0 +1,126 @@
...
+    <enum name="mode">
+      <entry name="off" value="0"
+             summary="Output is turned off."/>
+      <entry name="on" value="1"
+             summary="Output is turned on, no power saving"/>
+    </enum>
+
+    <enum name="error">
+      <entry name="invalid_mode" value="0"
+	     summary="the power save mode is not supported by the output"/>
```
emersion: Protocol errors should only be used when the client can avoid them. If the output doesn't support power management, we should just send the failed event without any protocol error.

unstable/wlr-output-power-management-unstable-v1.xml
```diff
@@ -0,0 +1,126 @@
...
+	mode a protocol error is raised.
+      </description>
+      <arg name="mode" type="uint" enum="mode" summary="the power save mode to set"/>
+    </request>
+
+    <event name="mode">
+      <description summary="Report a power save mode change">
+	Report the power save mode change of an output.
+
+	The mode event is sent after an output changed it's power save mode. The reason
```
emersion: Typo: its

unstable/wlr-output-power-management-unstable-v1.xml
```diff
@@ -0,0 +1,126 @@
...
+
+    <enum name="error">
+      <entry name="invalid_mode" value="0"
+	     summary="the power save mode is not supported by the output"/>
+    </enum>
+
+    <request name="set_mode">
+      <description summary="Set an outputs power save mode">
+	Set an output's power save mode to the given mode. The mode change
+	is effective immediately. If the output does not suppoert the given
```
emersion: Typo: support (but see comment about protocol errors)

unstable/wlr-output-power-management-unstable-v1.xml
agx: i think the code already does that https://github.com/agx/wlroots/blob/wlr-output-power-management/types/wlr_output_power_management_v1.c#L81 but i just forgot to update the protocol.

unstable/wlr-output-power-management-unstable-v1.xml
agx: fixed

agx:
> General style comment: could we avoid mixing tabs and spaces? 

done. the current branch is here:

https://github.com/agx/wlroots/blob/wlr-output-power-management/types/wlr_output_power_management_v1.c#L81

i'll do a mr once i got a chance to test this.

unstable/wlr-output-power-management-unstable-v1.xml
agx: i've added a `invalid_mode` event - hope that's what was intended here.

unstable/wlr-output-power-management-unstable-v1.xml
```diff
@@ -0,0 +1,128 @@
...
+        including:
+        - The output doesn't support power management
+        - Another client already has exclusive power management control
+          for this output
+
+        Upon receiving this event, the client should destroy this object.
+      </description>
+    </event>
+
+    <event name="invalid_mode">
```
emersion: This is racy -- if you issue two `set_mode` requests, then get an `invalid_mode` event, you don't know which mode isn't supported.

But I wonder if we need this event at all. Right now there's only on/off so the compositor can immediately send `failed` if the output doesn't support power management. It would be easy to add a new `supported_mode` event in the future to advertise a list of supported modes if we need it.

unstable/wlr-output-power-management-unstable-v1.xml
```diff
@@ -0,0 +1,128 @@
...
+        mode a failed event is sent.
+      </description>
+      <arg name="mode" type="uint" enum="mode" summary="the power save mode to set"/>
+    </request>
+
+    <event name="mode">
+      <description summary="Report a power save mode change">
+        Report the power save mode change of an output.
+
+        The mode event is sent after an output changed its power save mode. The reason
```
emersion: We should probably send it when the object is created too.

unstable/wlr-output-power-management-unstable-v1.xml
agx: o.k. let's drop that.

unstable/wlr-output-power-management-unstable-v1.xml
agx: with the dropped `invalid_mode` and a `failed` event this seems resolve too.

unstable/wlr-output-power-management-unstable-v1.xml
```diff
@@ -0,0 +1,123 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<protocol name="wlr_output_power_management_v1">
```
emersion: This should be named `wlr_output_power_management_unstable_v1`

unstable/wlr-output-power-management-unstable-v1.xml
```diff
@@ -0,0 +1,123 @@
...
+    backward incompatible changes may be made. Backward compatible changes
+    may be added together with the corresponding interface version bump.
+    Backward incompatible changes are done by bumping the version number in
+    the protocol and interface names and resetting the interface version.
+    Once the protocol is to be declared stable, the 'z' prefix and the
+    version number in the protocol and interface names are removed and the
+    interface version number is reset.
+  </description>
+
+  <interface name="zwlr_output_power_management_manager_v1" version="1">
```
emersion: Maybe we could just rename this interface to `zwlr_output_power_manager_v1`?

unstable/wlr-output-power-management-unstable-v1.xml
```diff
@@ -0,0 +1,123 @@
...
+        All objects created by the manager will still remain valid, until their
+        appropriate destroy request has been called.
+      </description>
+    </request>
+
+  </interface>
+  <interface name="zwlr_output_power_management_v1" version="1">
+    <description summary="adjust power save mode for an output">
+      This object is a manager which offers requests to set the power
+      save mode of outputs.
```
emersion: This description is for the manager. Should be updated for this per-output object.

unstable/wlr-output-power-management-unstable-v1.xml
```diff
@@ -0,0 +1,123 @@
...
+    </event>
+
+    <event name="failed">
+      <description summary="object no longer valid">
+        This event indicates that the output power management control
+        is no longer valid. This can happen for a number of reasons,
+        including:
+        - The output doesn't support power management
+        - Another client already has exclusive power management control
+          for this output
```
emersion: Maybe we could add to the list "the output disappeared"?

emersion:
This looks pretty good to me overall.

unstable/wlr-output-power-management-unstable-v1.xml
agx: fixed.

unstable/wlr-output-power-management-unstable-v1.xml
agx: i meant the object that manges the outputs power management  - hopefully it's clearer now.

unstable/wlr-output-power-management-unstable-v1.xml
agx: good point.

unstable/wlr-output-power-management-unstable-v1.xml
agx: making this short is better.

agx:
i've changed the text to consistently use 'output power management mode' - instead of a mixture of  'output power save mode' and 'output power management mode'. the later also seems more flexible for future additions.

unstable/wlr-output-power-management-unstable-v1.xml
```diff
@@ -0,0 +1,124 @@
...
+    interface version number is reset.
+  </description>
+
+  <interface name="zwlr_output_power_manager_v1" version="1">
+    <description summary="manager to create per-output power management">
+      This interface is a manager that allows creating per-output power
+      management mode controls.
+    </description>
+
+    <request name="get_output_power_management">
```
emersion: Because we renamed the objects, this can be renamed to `get_output_power` too.

unstable/wlr-output-power-management-unstable-v1.xml
```diff
@@ -0,0 +1,124 @@
...
+      <arg name="output" type="object" interface="wl_output"/>
+    </request>
+
+    <request name="destroy" type="destructor">
+      <description summary="destroy the manager">
+        All objects created by the manager will still remain valid, until their
+        appropriate destroy request has been called.
+      </description>
+    </request>
+
```
emersion: Style nit: move this newline rigth after `</interface>`

emersion:
LGTM apart from these nits, will review the wlroots impl soon.

unstable/wlr-output-power-management-unstable-v1.xml
agx: if found `get_output_power` a bit misleading but renaming is more consistent.

unstable/wlr-output-power-management-unstable-v1.xml
agx: done.

agx:
the wlroots side will need work after the protocol and naming changes but i wanted to have the protocol settle do avoid excessive renames. i'll give this a respin over the next days.

unstable/wlr-output-power-management-unstable-v1.xml
```diff
@@ -0,0 +1,124 @@
...
+    THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+    LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+    FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+    DEALINGS IN THE SOFTWARE.
+  </copyright>
+
+  <description summary="Control power management modes of outputs">
+    This protocol allows clients to control power management modes
+    of outputs that are currently part of the compositor space. The
+    intent is to allow speial clients like desktop shells to power
```
apreiml: Typo: speial

agx: fixed. thanks.

ddevault:
+1 from me as well, also want to see a wlr implementation

ddevault:
Thanks!

Issue #69, opened 2019-12-31
virtual-pointer: add a output hint


emersion:
Add an optional hint for the output the virtual pointer could be mapped to.

PR #71, opened 2020-02-05
virtual-pointer: Add request for mapping to specific output

any1:
Closes: https://github.com/swaywm/wlr-protocols/issues/69

### DIFF
--- a/unstable/wlr-virtual-pointer-unstable-v1.xml
+++ b/unstable/wlr-virtual-pointer-unstable-v1.xml
@@ -23,7 +23,7 @@
     DEALINGS IN THE SOFTWARE.
   </copyright>
 
-  <interface name="zwlr_virtual_pointer_v1" version="1">
+  <interface name="zwlr_virtual_pointer_v1" version="2">
     <description summary="virtual pointer">
       This protocol allows clients to emulate a physical pointer device. The
       requests are mostly mirror opposites of those specified in wl_pointer.
@@ -118,7 +118,7 @@
     </request>
   </interface>
 
-  <interface name="zwlr_virtual_pointer_manager_v1" version="1">
+  <interface name="zwlr_virtual_pointer_manager_v1" version="2">
     <description summary="virtual pointer manager">
       This object allows clients to create individual virtual pointer objects.
     </description>
@@ -134,6 +134,19 @@
 
     <request name="destroy" type="destructor" since="1">
       <description summary="destroy the virtual pointer manager"/>
+    </request>
+
+    <!-- Version 2 additions -->
+    <request name="create_virtual_pointer_with_output" since="2">
+      <description summary="Create a new virtual pointer">
+        Creates a new virtual pointer. The seat and the output arguments are
+        optional. If the seat argument is set, the compositor should assign the
+        input device to the requested seat. If the output argument is set, the
+        compositor should map the input device to the requested output.
+      </description>
+      <arg name="seat" type="object" interface="wl_seat" allow-null="true"/>
+      <arg name="output" type="object" interface="wl_output" allow-null="true"/>
+      <arg name="id" type="new_id" interface="zwlr_virtual_pointer_v1"/>
     </request>
   </interface>
 </protocol>
--- a/unstable/wlr-virtual-pointer-unstable-v1.xml
+++ b/unstable/wlr-virtual-pointer-unstable-v1.xml
@@ -29,6 +29,13 @@
       requests are mostly mirror opposites of those specified in wl_pointer.
     </description>
 
+    <enum name="error">
+      <entry name="invalid_axis" value="0"
+        summary="client sent invalid axis enumeration value" />
+      <entry name="invalid_axis_source" value="1"
+        summary="client sent invalid axis source enumeration value" />
+    </enum>
+
     <request name="motion">
       <description summary="pointer relative motion event">
         The pointer has moved by a relative amount to the previous request.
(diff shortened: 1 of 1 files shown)


unstable/wlr-virtual-pointer-unstable-v1.xml
```diff
@@ -111,7 +118,7 @@
     </request>
   </interface>
 
-  <interface name="zwlr_virtual_pointer_manager_v1" version="1">
+  <interface name="zwlr_virtual_pointer_manager_v1" version="2">
```
emersion: The version needs to be bumped on `zwlr_virtual_pointer_v1` too

unstable/wlr-virtual-pointer-unstable-v1.xml
```diff
@@ -29,6 +29,13 @@
       requests are mostly mirror opposites of those specified in wl_pointer.
     </description>
 
+    <enum name="error">
```
emersion: So the wlroots version and the upstream wlr-protocols version were out-of-sync, right?

unstable/wlr-virtual-pointer-unstable-v1.xml
```diff
@@ -124,5 +131,19 @@
       <arg name="seat" type="object" interface="wl_seat" allow-null="true"/>
       <arg name="id" type="new_id" interface="zwlr_virtual_pointer_v1"/>
     </request>
+    <request name="destroy" type="destructor" since="1">
```
emersion: Ditto?

unstable/wlr-virtual-pointer-unstable-v1.xml
```diff
@@ -124,5 +131,19 @@
...
+      <description summary="destroy the virtual pointer manager"/>
+    </request>
+
+    <request name="create_virtual_pointer2" since="2">
+      <description summary="Create a new virtual pointer">
+        Creates a new virtual pointer. The seat and the output arguments are
+        optional. If they are set, the compositor must bind the pointer
```
emersion: -1 to "must". These are hints the compositor is free to ignore.

unstable/wlr-virtual-pointer-unstable-v1.xml
```diff
@@ -124,5 +131,19 @@
       <arg name="seat" type="object" interface="wl_seat" allow-null="true"/>
       <arg name="id" type="new_id" interface="zwlr_virtual_pointer_v1"/>
     </request>
+    <request name="destroy" type="destructor" since="1">
+      <description summary="destroy the virtual pointer manager"/>
+    </request>
+
+    <request name="create_virtual_pointer2" since="2">
```
emersion: Maybe name this `create_virtual_pointer_with_output`?

emersion: A convention is to add a `<!-- Version 2 additions -->` comment before the new requests/events so that we make sure not to add anything out-of-order.

unstable/wlr-virtual-pointer-unstable-v1.xml
any1: So I write software that _maybe_ works based on the whims of the compositor?

unstable/wlr-virtual-pointer-unstable-v1.xml
any1: That appears to be the case. I should put the realignment into a separate commit.

unstable/wlr-virtual-pointer-unstable-v1.xml
ascent12: Most of the Wayland protocols follow the RFC 2119 terms, so "should" would be a good replacement word.

>So I write software that maybe works based on the whims of the compositor?

Wayland in a nutshell.

---

As someone who knows very little about this protocol and don't have a lot of context for it, I find this description extremely vague.

any1: >>So I write software that maybe works based on the whims of the compositor?

>Wayland in a nutshell.

It can't be all that bad. In any case I'd like to make things better in that regard, not worse. If the desired behaviour is not guaranteed, there needs to be a way for the client to discover that the compositor decided not to follow its advice so that it can take the appropriate action to avert disaster. I think it's much easier to put strong guarantees into the standard than to add a result event that tells the client about what actually happened.

ddevault: For what it's worth, it's possible for the protocol to place constraints on the compositor's behavior. The compositor doesn't need to follow them - and really, neither does the client - and they can accept the technical and social fallout which ensues. Consider for example that fullscreen clients are requried not to be compositied, which has been a point of contention for us on sway for a while - the more user-friendly option may be to not composite them, but the protocol says so therefore it's right has been the party line for a while (yeah, yeah, there's more nuance to that example, but it's just an example).

any1: I changed the wording to "should". Will that suffice?

emersion: Yes, the compositor will always have the last word. The compositor could already ignore requests sent via virtual-pointer, e.g. if the client isn't authorized to inject events or if the compositor offers a way for users to enable/disable event injection.

The "should" wording LGTM.

One other way to fix this would be to add a "failed" event to the pointer interface to let clients know that the virtual pointer is no longer active. That way, compositors could send the event if  they refuse to create the virtual pointer, if the user has decided to shut down the virtual pointer, or if the compositor refuses to map the pointer to the specified output.

Usually, we stick with the "should" wording so that the compositor can override the client's hints. In this particular case, I don't have a good use-case in mind: not mapping the virtual pointer to the specified output will just result in unexpected behavior. So I'd be fine with the other approach.

any1: So, the change is good enough as is?

emersion: Yes.

emersion:
Protocol looks good. Will review the wlroots patch before merging this PR.

emersion:
Thanks!

PR #72, opened 2020-02-08
WiP: Add a seat management protocol

any1:
This is a new protocol for seat management from within wayland clients.

See: https://github.com/any1/wayvnc/issues/3


emersion:
Superseded by https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/61

PR #75, opened 2020-03-27
Add a sync_fence event to the export-dmabuf protocol.

cyanreg:
This commit adds a sync_fence FD to the export-dmabuf protocol, allowing
for clients to synchronize their readout of the frame with the last
operation the compositor performs to the frame, thus eliminating any
race conditions where the compositor gives the frame to a client while
still compositing to it.

### DIFF
--- a/unstable/wlr-export-dmabuf-unstable-v1.xml
+++ b/unstable/wlr-export-dmabuf-unstable-v1.xml
@@ -36,7 +36,7 @@
     interface version number is reset.
   </description>
 
-  <interface name="zwlr_export_dmabuf_manager_v1" version="1">
+  <interface name="zwlr_export_dmabuf_manager_v1" version="2">
     <description summary="manager to inform clients and begin capturing">
       This object is a manager with which to start capturing from sources.
     </description>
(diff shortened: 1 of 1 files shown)


unstable/wlr-export-dmabuf-unstable-v1.xml
```diff
@@ -139,11 +142,22 @@
            summary="index of the the plane the data in the object applies to"/>
     </event>
 
+    <event name="sync_fence" since="2">
```
emersion: We can't add an event in the middle of the protocol, because the event opcode is its index. We generally add news events at the end of the interface, after `<!-- Version 2 additions -->`.

unstable/wlr-export-dmabuf-unstable-v1.xml
```diff
@@ -59,7 +59,7 @@
     </request>
   </interface>
 
-  <interface name="zwlr_export_dmabuf_frame_v1" version="1">
+  <interface name="zwlr_export_dmabuf_frame_v1" version="2">
```
emersion: The minor version of the `zwlr_export_dmabuf_manager_v1` interface needs to be bumped as well (since versions are inherited from the parent object).

emersion (commented):
Apart from these comments, this patch LGTM. I'll add a commit to my explicit-sync wlroots branch to add support for this.

emersion:

https://gitlab.freedesktop.org/wlroots/wlr-protocols/-/merge_requests/75

Issue #73, opened 2020-03-24
Add DMA-BUF support to screencopy


emersion:
This would perform a GPU copy instead of a CPU one. screencopy already uses `wl_buffer` so it shouldn't be too hard.

- Indicate which device to allocate buffers on
- Indicate the modifier
- (Future  issue) How does that play with explicit sync?

emersion:
Implementation detail: on the compositor side, either of these could be used:

* https://www.khronos.org/registry/OpenGL/extensions/ANGLE/ANGLE_framebuffer_blit.txt (https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/3149)
* https://www.khronos.org/registry/OpenGL/extensions/NV/NV_framebuffer_blit.txt

ammen99:
Out of curiosity, why would we want this instead of the export-dmabuf approach?

ascent12:
I consider export-dmabuf to be borderline broken, and has some pretty bad race conditions. While copy-to-dmabuf isn't perfectly efficient, it'll still be pretty good, and I think it's much more preferable.

emersion:
I disagree. Renderer v6 should allow us to mitigate race conditions in export-dmabuf.

ascent12:
By throwing out any allocations once they've been exported?

emersion:
No, by adding a "release" request and allowing the client to keep a reasonable number of DMA-BUFs locked.

PR #80, opened 2020-05-21
screencopy: Extend protocol to allow linux-dmabuf

any1:
Protocol changes for https://github.com/swaywm/wlroots/pull/2133

Closes https://github.com/swaywm/wlr-protocols/issues/73

### DIFF
--- a/unstable/wlr-screencopy-unstable-v1.xml
+++ b/unstable/wlr-screencopy-unstable-v1.xml
@@ -38,7 +38,7 @@
     interface version number is reset.
   </description>
 
-  <interface name="zwlr_screencopy_manager_v1" version="2">
+  <interface name="zwlr_screencopy_manager_v1" version="3">
     <description summary="manager to inform clients and begin capturing">
       This object is a manager which offers requests to start capturing from a
       source.
@@ -80,13 +80,18 @@
     </request>
   </interface>
 
-  <interface name="zwlr_screencopy_frame_v1" version="2">
+  <interface name="zwlr_screencopy_frame_v1" version="3">
     <description summary="a frame ready for copy">
       This object represents a single frame.
 
-      When created, a "buffer" event will be sent. The client will then be able
-      to send a "copy" request. If the capture is successful, the compositor
-      will send a "flags" followed by a "ready" event.
+      When created, a series of buffer events will be sent, each representing a
+      supported buffer type. The "buffer_done" event is sent afterwards to
+      indicate that all supported buffer types have been enumerated. The client
+      will then be able to send a "copy" request. If the capture is successful,
+      the compositor will send a "flags" followed by a "ready" event.
+
+      For objects version 2 or lower, wl_shm buffers are always supported, ie.
+      the "buffer" event is guaranteed to be sent.
 
       If the capture failed, the "failed" event is sent. This can happen anytime
       before the "ready" event.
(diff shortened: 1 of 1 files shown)


unstable/wlr-screencopy-unstable-v1.xml
```diff
@@ -102,6 +102,12 @@
 
         The client should then create a buffer with the provided attributes, and
         send a "copy" request.
+
+        When the buffer is a dmabuf, the stride parameter may be ignored.
+
+        The format parameter shall be interpreted according to
```
emersion, line 108: This part should be covered by https://github.com/swaywm/wlr-protocols/pull/81

emersion:
From a protocol point of view, if we didn't have to maintain backwards compat, I think it would be better to have one event per supported buffer type, each with buffer parameters, followed by a "done" event (which would mark the end of buffer param events).

This would allow users to know exactly which types of buffers are supported. This would avoid having to shoehorn all buffer types params into a single event: `wl_shm.format` shouldn't be used with linux-dmabuf and `stride` may be ignored for linux-dmabuf.

any1:
Updated according to review.

emersion:
I'd prefer having one event per buffer type advertising the buffer parameters.

any1:
> I'd prefer having one event per buffer type advertising the buffer parameters.

Ok, committed something to that effect just now.

unstable/wlr-screencopy-unstable-v1.xml
```diff
@@ -203,5 +203,18 @@
       <arg name="width" type="uint" summary="current width"/>
       <arg name="height" type="uint" summary="current height"/>
     </event>
+
+    <!-- Version 3 additions -->
+    <event name="linux_dmabuf" since="3">
+      <description summary="linux-dmabuf buffer information">
+        Provides information about the frame's linux-dmabuf buffer.
+
+        This event is sent once before the regular "buffer" event if
```
emersion, line 212: Maybe we should add a `buffer_done` event instead relying on `buffer`?

emersion:
Descriptions of other requests mention the `buffer` event. They need to be updated.

unstable/wlr-screencopy-unstable-v1.xml
any1: Yes, that would make it possible for compositors which implement the protocol to omit `wl_shm`.

any1:
Updated.

emersion:
The new events look good. Please update the `zwlr_screencopy_frame_v1` and `copy` descriptions.

any1:
> The new events look good. Please update the `zwlr_screencopy_frame_v1` and `copy` descriptions.

Done.

unstable/wlr-screencopy-unstable-v1.xml
```diff
@@ -80,13 +80,15 @@
...
       This object represents a single frame.
 
-      When created, a "buffer" event will be sent. The client will then be able
-      to send a "copy" request. If the capture is successful, the compositor
-      will send a "flags" followed by a "ready" event.
+      When created, a series of buffer events will be sent, each representing a
+      supported buffer type. The "buffer_done" event is sent afterwards to
+      indicate that all supported buffer types have been enumerated. The client
+      will then be able to send a "copy" request. If the capture is successful,
+      the compositor will send a "flags" followed by a "ready" event.
```
emersion, line 91: Ah, can we just add a clarification here: "For objects version 2 or lower, wl_shm buffers are always supported, ie. the "buffer" event is guaranteed to be sent."

any1: Done

emersion:
Thanks!

PR #83, opened 2020-06-03
screencopy: clarify the buffer event description

emersion:
The buffer event provides information specific to wl_shm buffers only.
Without these clarifications, a client can only guess the supported
buffer types and may receive a protocol error when its guess is wrong.

cc @any1


emersion:
Superseded by https://github.com/swaywm/wlr-protocols/pull/80

PR #86, opened 2020-06-23
layer-shell: introduce ack for new outputs

ifreund:
This ack sequence eliminates the race between a client creating a new layer surface and the compositor rendering the first frame of a new output.

This race is especially important for screenlockers using layer-shell such as swaylock https://github.com/swaywm/swaylock/issues/16

### DIFF
--- a/unstable/wlr-layer-shell-unstable-v1.xml
+++ b/unstable/wlr-layer-shell-unstable-v1.xml
@@ -25,7 +25,7 @@
     THIS SOFTWARE.
   </copyright>
 
-  <interface name="zwlr_layer_shell_v1" version="3">
+  <interface name="zwlr_layer_shell_v1" version="4">
     <description summary="create surfaces that are layers of the desktop">
       Clients can use this interface to assign the surface_layer role to
       wl_surfaces. Such surfaces are assigned to a "layer" of the output and
@@ -92,9 +92,43 @@
         are not affected.
       </description>
     </request>
+
+    <!-- Version 4 additions -->
+
+    <event name="new_output" since="4">
+      <description summary="notify of a new wl_output">
+        This event indicates that one or more new outputs have been
+        created. All clients must send an ack_new_output request in response
+        to this event. If a client wishes to create a new layer surface in
+        response to the new output(s), it must make an ack_new_output request
+        with the provided serial after creating any new layer surfaces using
+        the get_layer_surface request and before the first commit on any of
+        the new surfaces. If the client does not wish to create a new layer
+        surface, it should ack immediately.
+
+        This allows the compositor to wait until new layer surfaces are ready
+        before rendering the first frame of new outputs, avoiding a race that
+        could cause imperfect frames.
+      </description>
+      <arg name="serial" type="uint"/>
+    </event>
+
+    <request name="ack_new_output" since="4">
+      <description summary="ack a new_output event">
+        This request informs the server that the client has created all
+        layer surfaces it intends to create in response to the new output(s).
+        These new layer surfaces are associated only with this ack sequence
+        and do *not* carry over to following ack sequences should any occur
+        before the first commit on the surfaces.
+
+        If the client receives multiple new_output events before it
+        can respond to one, it only has to ack the last new_output event.
+      </description>
+      <arg name="serial" type="uint" summary="serial from a new_output event"/>
+    </request>
   </interface>
 
-  <interface name="zwlr_layer_surface_v1" version="3">
+  <interface name="zwlr_layer_surface_v1" version="4">
     <description summary="layer metadata interface">
       An interface that may be implemented by a wl_surface, for surfaces that
       are designed to be rendered as a layer of a stacked desktop-like


emersion:
If wonder if it would be desirable to change this to a ping/pong mechanism [similar to xdg-shell](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/blob/master/stable/xdg-shell/xdg-shell.xml#L92).

unstable/wlr-layer-shell-unstable-v1.xml
```diff
@@ -82,9 +82,37 @@
...
       <entry name="overlay" value="3"/>
     </enum>
+
+    <!-- Version 3 additions -->
+
+    <event name="new_output" since="3">
+      <description summary="notify of a new wl_output">
+        This event indicates that a new output has been created. If the client
+        wishes to create a new layer surface in response to this new output,
+        it must make an ack_new_output request with the provided serial
```
emersion, line 92: If the client doesn't want to create a new layer surface, it still needs to send the ack.

ifreund:
Clarified that all clients must ack and rebased onto #87 to avoid conflicts.

> If wonder if it would be desirable to change this to a ping/pong mechanism similar to xdg-shell.

If I understand correctly, this would just make the ability to use this to achieve atomicity of new outputs implicit instead of explicit. The logic would stay the same from the server's point of view:

1. post new output global
3. send ping to all layer-shells
3. if a new layer surface is requested on the new output before receiving a pong from that layer_shell, wait for a committed buffer before enabling/drawing the first frame of the new output.

This seems like it would work just as well and could be more generally useful than the new output specific language.

ifreund:
The disadvantage to solving this race implicitly through a ping/pong is that clients which for whatever reason make the pong request before get_layer_surface will still be racy. We could certainly explain how compositors could use ping/pong to solve this in the protocol and encourage clients to respond to the events in order though.

ifreund:
Opened up  #90 to explore the ping/pong alternative approach.

emersion:
>If I understand correctly, this would just make the ability to use this to achieve atomicity of new outputs implicit instead of explicit.

Yeah. That would work similarly to `wl_display_roundtrip`: the ping-pong roundtrip guarantees that the client has processed all events until the ping request.

This also has more use-cases, like detecting unresponsive layer-shell clients (like xdg-shell).

unstable/wlr-layer-shell-unstable-v1.xml
```diff
@@ -92,6 +92,35 @@
         are not affected.
       </description>
     </request>
+
+    <event name="new_output" since="3">
+      <description summary="notify of a new wl_output">
+        This event indicates that a new output has been created. All clients
```
emersion, line 98: It would be nice to allow multiple outputs to be advertised before sending this event. (Could be useful when plugging a DisplayPort MST hub for instance.)

ifreund: Agreed, I'll add that to the description.

ifreund:
Tweaked the descriptions to clarify that the compositor may create several new outputs at once before sending this event. Also bumped the protocol version to 4 as there was a wlroots release with version 3 since this PR was opened.

unstable/wlr-layer-shell-unstable-v1.xml
```diff
@@ -92,9 +92,41 @@
...
+      <arg name="serial" type="uint"/>
+    </event>
+
+    <request name="ack_new_output" since="3">
+      <description summary="ack a new_output event">
+        This request informs the server that following commits on layer surfaces
+        created by the client take the new output(s) into account.
+
+        A client may make multiple ack_new_output requests before committing.
+        The last request made before the commit indicates which new_output
```
ammen99, line 122: Sorry if this is a stupid question, but: what does this last sentence refer to? Aren't next commits taking into account all acked outputs?

ifreund: The idea with this paragraph was to parallel the configure/ack behavior of the xdg-shell protocol. However, I think it may be over-complicating things. There are two paths we could take here:

1. State that new layer surfaces created on a new output before acking that new output are associated only with that change of state.
2. In the case that a second new output is created before the first commit on a new layer surface which was created in response to a previous `new_output` event and the client acks the second new output before committing the layer surface, the layer surface then becomes part of the ack sequence for the second new output.

In it's current state this paragraph is, I believe, trying to explain option 2. I'd suggest that we switch this over to option 1 though as it's simpler and I believe more ideal behavior. Creation of a second new output wouldn't prolong the waiting period for the first frame on a prior new output.

Hopefully that all makes sense, it's rather late here and I don't feel like I've done a terribly good job of putting it into words.

ammen99: IMHO 1. makes more sense

unstable/wlr-layer-shell-unstable-v1.xml
```diff
@@ -92,9 +92,40 @@
...
+      <arg name="serial" type="uint"/>
+    </event>
+
+    <request name="ack_new_output" since="4">
+      <description summary="ack a new_output event">
+        This request informs the server that the client has created all
+        layer surfaces it intends to create in response to the new output(s).
+        These new layer surfaces are associated only with this ack sequence
+        and do *not* carry over to following ack sequences should any occur
+        before the first commit on the surfaces.
```
emersion, line 122: Should we allow clients to only ack the latest `new_output` serial (like xdg-shell does)? Basically:

```
<- new_output(serial=42)
<- new_output(serial=43)
ack_new_output(serial=43)
```

The ack with serial=43 implicitly acks the new_output event with serial=42. This allows client implementations to just keep a single serial rather than a list of serials.

emersion:
Overall looks good. Would be nice to see a client & compositor implementation!

unstable/wlr-layer-shell-unstable-v1.xml
ifreund: Good call, done.

emersion (commented):
LGTM

emersion:

https://gitlab.freedesktop.org/wlroots/wlr-protocols/-/merge_requests/86

PR #87, opened 2020-06-24
layer-shell: add destroy request

ifreund:
~Depends on #86 to avoid conflicts.~


emersion:
Since this is a less complicated change, can we make this go in before #86?

ifreund:
Sure, I'll rebase this the other way around and send a wlroots implementation for the destroy request since it's fairly trivial.

ifreund:
Implemented in https://github.com/swaywm/wlroots/pull/2308

Issue #88, opened 2020-06-30
Formally reject new protocols


emersion:
New protocols should go in wayland-protocols' ext namespace, which has no more inclusion requirements than this repo, but also offers reviews by non-wlroots people and allows to host all protocols in the same place.

Proposal: reject new protocols in wlr-protocols. Only accept updates for protocols not yet included in wayland-protocols.

cc @ddevault

agx:
This sounds good - maybe it would be good to mention how to do that in the `README.md` of this repo?

ddevault:
ack

PR #89, opened 2020-06-30
Add README

emersion:
Closes: https://github.com/swaywm/wlr-protocols/issues/88


ammen99:
Does this also apply to all open PRs ?

emersion:
I'd say yes, since there's no good reason to continue to push new protocols to wlr-protocols. The inclusion requirements are more or less the same for wayland-protocols ext namespace, so this shouldn't change much for contributors.

PR #94, opened 2020-08-13
Add serial to data control to avoid race condition with clipboard managers

davidedmundson:
A typically clipboard manager works as follows:

 - On a new clipboard, we copy any text
- If the clipboard becomes empty (typically because a client died), we
paste our previously copied text

Our race happens as follows:

first copy:
- client creates a wl_data_offer
- compositors sees it forwards it clipboard manager which copies the
text...

second copy:
 - client deletes its old wl_data_offer before creating a new one

 - compositor forwards this update to clipboard manager

- clipboard manager knows the clipboard is empty and starts its
operation to prevent an empty clipboard

 - client creates a new wl_data_offer and calls set_selection

- clipboard manager  creates a new wlr_data_offer (with the old
clipboard text) and calls set_selection

The compositor can get these last two in any order, and we end up
replacing our new clipboard content with out-of-date previous clipboard
contents.

This patch adds a serial number that can be used when a set_selection is
used in repsponse to a selection event.

### DIFF
--- a/unstable/wlr-data-control-unstable-v1.xml
+++ b/unstable/wlr-data-control-unstable-v1.xml
@@ -41,7 +41,7 @@
     interface version number is reset.
   </description>
 
-  <interface name="zwlr_data_control_manager_v1" version="2">
+  <interface name="zwlr_data_control_manager_v1" version="3">
     <description summary="manager to control data devices">
       This interface is a manager that allows creating per-seat data device
       controls.
@@ -71,7 +71,7 @@
     </request>
   </interface>
 
-  <interface name="zwlr_data_control_device_v1" version="2">
+  <interface name="zwlr_data_control_device_v1" version="3">
     <description summary="manage a data device for a seat">
       This interface allows a client to manage a seat's selection.
 
@@ -141,7 +141,7 @@
 
     <!-- Version 2 additions -->
 
-    <event name="primary_selection" since="2">
+    <event name="primary_selection" since="3">
       <description summary="advertise new primary selection">
         The primary_selection event is sent out to notify the client of a new
         wlr_data_control_offer for the primary selection for this device. The
@@ -183,9 +183,33 @@
       <entry name="used_source" value="1"
         summary="source given to set_selection or set_primary_selection was already used before"/>
     </enum>
+
+    <!-- Version 3 additions -->
+
+    <event name="selection_serial" since="3">
+      <description summary="copy data to the selection">
+        This should be sent by compositors before a selection event with an
+        increasing serial number.
+
+        This can be used by the client in partnership to set_selection_response
+        to indicate pairing of a selection event to a reply.
+      </description>
+      <arg name="serial" type="uint"/>
+    </event>
+
+    <event name="primary_selection_serial" since="3">
+      <description summary="copy data to the selection">
+        This should be sent by compositors before a primary_selection event
+        with an increasing serial number.
+
+        This can be used by the client in partnership to set_selection_response
+        to indicate pairing of a selection event to a reply.
+      </description>
+      <arg name="serial" type="uint"/>
+    </event>
   </interface>
 
-  <interface name="zwlr_data_control_source_v1" version="1">
+  <interface name="zwlr_data_control_source_v1" version="3">
     <description summary="offer to transfer data">
       The wlr_data_control_source object is the source side of a
       wlr_data_control_offer. It is created by the source client in a data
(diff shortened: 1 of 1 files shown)


unstable/wlr-data-control-unstable-v1.xml
```diff
@@ -183,6 +183,74 @@
...
+
+    <event name="selection_serial" since="3">
+      <description summary="copy data to the selection">
+        This should be sent by compositors before a selection event with an
+        increasing serial number.
+
+        This can be used by the client in partnership to set_selection_response to indicate
+        pairing of a selection event to a reply.
+      </description>
+      <arg name="serial" type="uint">
```
zzag, line 197: The arg tag is not closed.

zzag (commented):
It looks like you forgot to bump the version of the zwlr_data_control_device_v1 interface. ;-)

unstable/wlr-data-control-unstable-v1.xml
emersion: Yeah, that's probably what causes the CI failure.

emersion:
The commit message (and PR title) seems broken. Copy-pasta gone wrong? :P

unstable/wlr-data-control-unstable-v1.xml
```diff
@@ -183,6 +183,74 @@
...
+    </event>
+
+    <request name="set_selection_response" since="3">
```
emersion, line 200: Bikeshedding: maybe we could find a better name? Would `swap_selection` make sense, since it replaces the selection with the provided serial? Or `set_selection_with_serial`?

emersion (commented):
Apart from @zzag's comments, looks good! Will wait for a compositor and client impl before merging this.

davidedmundson:
>Will wait for a compositor and client impl before merging this.

Ack, I was hoping to get this approved in principle. Then write the code, then merge simultaneously (providing we don't find anything)

unstable/wlr-data-control-unstable-v1.xml
davidedmundson: Generally speaking in my use case I'm trying to go from being empty to having content. So "swap" seems slightly off.

set_selection_with_serial is a very literal approach but that works for me.

davidedmundson: Throwing out another alternative - just so we consider all the options. We could add the response serial to the data control source. It would be an optional thing to send there and avoid having two requests here.

davidedmundson:
There is now an implementation in our library which passes a unit tests made to recreate this issue.

emersion:
Cool! Can you add a link to the server and client implementations?

davidedmundson:
(updated with word wrapping)

server https://invent.kde.org/plasma/kwayland-server/-/merge_requests/75
client https://invent.kde.org/plasma/plasma-workspace/-/merge_requests/232/

bugaevc:
Hi @davidedmundson!

It's great that you're trying to work on protocol improvements for fixing this race. I've need thinking about issues like the one you're running into, and more, for quite a while. (Only thing I don't understand is how come, if this has been pending for weeks, @emersion told me just yesterday that "There's no need for any new protocol extension/update"& "the KDE clipboad manager folks don't need anything else".)

Adding some sort of serial even would sure help fix *some* races, but unfortunately there are more/deeper issues here. Please read my previous thoughts on the matter ([in this discussion](https://github.com/bugaevc/wl-clipboard/issues/65#issuecomment-533781561) and [here](https://github.com/swaywm/sway/issues/5480#issuecomment-687835117)). I've [previously proposed](https://github.com/bugaevc/wl-clipboard/issues/65#issuecomment-533790918) proposed an API like this:

```xml
    <request name="set_backup_source" since="3">
      <description summary="set a backup selection source">
        This request asks the compositor to set the given source as the *backup* source
        for the given selection offer that is coming from another client.

        If and when the source backing the offer abruptly goes away while being the active
        selection, the compositor should atomically replace the selection with the provided
        backup source, and send out the new wl_data_device.offer events to clients describing
        this backup source. The client can then expect to receive
        zwlr_data_control_source_v1.send and zwlr_data_control_source_v1.cancelled events
        as usual.

        If, on the other hand, the selection is later explicitly set to another source
        (or to nil), the compositor should cancel the provided backup source by means of an
        zwlr_data_control_source_v1.cancelled event.
      </description>
      <arg name="backup_source" type="object" interface="zwlr_data_control_source_v1" />
      <arg name="offer" type="object" interface="zwlr_data_control_offer_v1" />
    </request>
```

Note that an API like this would fix both the races / atomicity concerns (the destroyed source gets atomically replaced with the backup source by the compositor, without round-tripping to the clipboard manager, and without other clients ever seeing the intermediary `selection(nil)` state), and the *password manager vs dying client* problem, because the compositor would only use the backup source in the latter case, and cancel it immediately in the former case.

How does that sound? Naming, description, and everything else is up for debate/bikeshedding, of course.

P.S. I would appreciate if I was cc'd on discussions about this & related topics in the future :slightly_smiling_face:

emersion:
Please open a new issue for this, so that this PR's discussion stays readable.

unstable/wlr-data-control-unstable-v1.xml
emersion: That sounds like a good idea! Sources cannot be re-used anyways.

emersion: @davidedmundson, would you be willing to update the implementations if I update the protocol here? Would be nice to move forward with this PR. :)

subdiff: @davidedmundson ping, or can you directly update the protocol here too?

davidedmundson: >would you be willing to update the implementations if I update the protocol here? Would be nice to move forward with this PR. :)

I can update this too. Apologies for the delay. It's now in my immediate TODO queue.

emersion:

https://gitlab.freedesktop.org/wlroots/wlr-protocols/-/merge_requests/94

PR #97, opened 2020-08-25
layer-shell: Require an initial commit

zzag:
Currently, the compositor doesn't have enough information to decide if a
given unmapped layer surface can be rearranged or rather configured.

In order to address the ambiguity, this change adjusts the protocol spec
saying that every client must perform an initial commit without a buffer
attached after creating and setting up a layer surface.

If the compositor knows that the initial state for a specified surface
has not been committed yet, it can skip such a surface when re-arranging
layer shell clients.


emersion (approved):
The wording comes from xdg-shell.

Looks good to me, thanks!

(This is not a breaking change but a clarification.)

PR #98, opened 2020-09-03
output-management-v1: add head identifying events

### LINKED ISSUE #68: output-management: add make, model, serial
These would be useful for utilities matching make/model/serial against a configuration file, to recognize a screen even if it doesn't use the same connector. These would be optional (not all backends support it) and not intended to be displayed to the user (see "description").

subdiff:
The following information through separate events are added:
- make
- model
- serial_number

This should allow clients to identify a display over different sessions and
load configuration data back.

A note is added that the description should be preferred when representing a
display in UI to users but as a short form for example the model could be used
in this case of course too.

Related: https://github.com/swaywm/wlroots/pull/2392

Closes #68

### DIFF
--- a/unstable/wlr-output-management-unstable-v1.xml
+++ b/unstable/wlr-output-management-unstable-v1.xml
@@ -39,7 +39,7 @@
     interface version number is reset.
   </description>
 
-  <interface name="zwlr_output_manager_v1" version="1">
+  <interface name="zwlr_output_manager_v1" version="2">
     <description summary="output device configuration manager">
       This interface is a manager that allows reading and writing the current
       output device configuration.
@@ -125,7 +125,7 @@
     </event>
   </interface>
 
-  <interface name="zwlr_output_head_v1" version="1">
+  <interface name="zwlr_output_head_v1" version="2">
     <description summary="output device">
       A head is an output device. The difference between a wl_output object and
       a head is that heads are advertised even if they are turned off. A head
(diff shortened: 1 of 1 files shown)


unstable/wlr-output-management-unstable-v1.xml
```diff
@@ -125,7 +125,7 @@
     </event>
   </interface>
 
-  <interface name="zwlr_output_head_v1" version="1">
+  <interface name="zwlr_output_head_v1" version="2">
```
emersion, line 128: We need to bump the version of all interfaces in the protocol.

unstable/wlr-output-management-unstable-v1.xml
```diff
@@ -257,6 +257,55 @@
...
+      </description>
+      <arg name="model" type="string"/>
+    </event>
+
+    <event name="serial_number" since="2">
+      <description summary="head serial number">
+        This event describes the serial number of the head.
+
+        Together with the make and model events the purpose is to allow clients
+        to recognize heads from previous sessions and for example load load
```
emersion, line 300: Typo: double "load"

unstable/wlr-output-management-unstable-v1.xml
```diff
@@ -257,6 +257,55 @@
         resources associated with it.
       </description>
     </event>
+
+    <event name="make" since="2">
```
emersion, line 261: Can we add a `<!-- Version 2 additions -->` comment right above?

emersion (commented):
Overall this looks good apart from these nits!

unstable/wlr-output-management-unstable-v1.xml
emersion, line 261: Sometimes this information is unavailable (missing from the EDID, or virtual output). Maybe we should make the event optional?

subdiff: I would have said in this case it's just not sent. Do you mean that? So I should mention it in the description?

emersion: Yes!

subdiff: I added some text about it to the description.  Check it out if you see it fitting. There is a "strong" statement about guaranteed uniqueness of head identification which I think makes sense when one has make, model and serial number but one should take one moment to think about corner cases.

emersion: It's not uncommon to have broken EDIDs with the same (make, model, serial) tuple for different monitors.

unstable/wlr-output-management-unstable-v1.xml
```diff
@@ -257,9 +257,80 @@
...
+        identify the head by available information from other events but should
+        be aware that there is no guarantee of unique identification anymore.
+
+        It is not recommended to display the model string in UI to users. For
+        that the string provided by the description event should be preferred.
+      </description>
+      <arg name="model" type="string"/>
+    </event>
+
+    <event name="serial_number" since="2">
```
emersion, line 310: Nit: the serial isn't necessarily a number

subdiff: That's why the argument is a string. Shall we call the event just serial?

subdiff: But at least Wikipedia defines the notion serial number to not necessarily be a number: https://en.wikipedia.org/wiki/Serial_number

unstable/wlr-output-management-unstable-v1.xml
subdiff: How about:

> It is not guaranteed this event will be ever sent. A reason for that can be that the compositor does not have information about the make of the head or the definition of a make is not sensible in the current setup, for example in a virtual session. Clients can still try to identify the head by available information from other events but should be aware that there is an increased risk of false positives.

emersion: LGTM!

emersion (approved):
LGTM, thanks @subdiff!

Issue #32, opened 2018-12-15
Enhance wlr-layer-shell keyboard_interactivity


ammen99:
The current `keyboard_interactivity` allows clients to only notify the compositor they want exclusive keyboard access. It would make much more sense to allow clients to specify they want one of the following:

1) No keyboard interaction - backgrounds, some docks, etc
2) keyboard-focus-on-click - such could be panels that have a search field or that can be navigated with they keyboard, but don't want to lock focus to their layer
3) exclusive keyboard focus - like the current `1` value - can be used for ex. by lockscreens

Note that adding this doesn't have to break existing clients - the protocol currently says what happens when the value is `1` (and afaik it is implicitly understood the only other reasonable value is `0`). We can just specify what happens when the value is `2` (focus-on-click)

ddevault:
Can you elaborate more on some of the clients in the second group?

ammen99:
> Can you elaborate more on some of the clients in the second group?

I'm imagining something like a panel which has a search field (search for applications, files, whatever you want). Kind of what Windows has, although there the search field is inside the "popup". So, one may click on the panel's search field, and if he/she wants to dismiss it, can still click on the other normal apps.

Another use-case I thought of but as emersion pointed out it can be done via xdg-popup grab: navigating a popup menu with the keyboard.

And last but not least (a dirty hack, I agree) would be this: have a panel which can activate its menu from the terminal, via dbus or whatever. You can register this as a keybinding, then the panel first sets exclusive interactivity (stealing focus) then sets normal interactivity. At this point, the panel can be navigated with the keyboard, but it can also be dismissed when clicking on another window.

There aren't any clients that require this and that I know of, but I still think that differentiating between those 3 different cases is something that might help in the future.

emersion:
>keyboard-focus-on-click

Thought about it for a while and I'm not sure it fixes the issue. If you have a panel with a search field, you want the panel to receive focus when the user clicks on the search field but you don't want to receive focus when the user clicks anywhere else on the panel.

Maybe that can be fixed by adding a request to start a grab on the layer surface.

ammen99:
> Maybe that can be fixed by adding a request to start a grab on the layer surface.

Ok, that is also a good idea. But I think that anyway we need to clarify what happens with layer surfaces without keyboard interactivity - they shouldn't get keyboard focus at all, except by this grab request?

emersion:
[...]

PR #100, opened 2020-11-25
Proposal for more options regarding keyboard interactivity for layer-shell clients

dkondor:
I'm opening this as a concrete proposal for #32 

I believe the current protocol makes it hard to have layer-shell surfaces in the top layer (e.g. panels and docks) that have keyboard interaction on their main surface and not in popups.

Let me know if I should provide more motivation or use cases.

My intention for this is to be backward-compatible with the current protocol (0/1 values for keyboard interactivity keep their meaning).

### DIFF
--- a/unstable/wlr-layer-shell-unstable-v1.xml
+++ b/unstable/wlr-layer-shell-unstable-v1.xml
@@ -25,7 +25,7 @@
     THIS SOFTWARE.
   </copyright>
 
-  <interface name="zwlr_layer_shell_v1" version="3">
+  <interface name="zwlr_layer_shell_v1" version="4">
     <description summary="create surfaces that are layers of the desktop">
       Clients can use this interface to assign the surface_layer role to
       wl_surfaces. Such surfaces are assigned to a "layer" of the output and
@@ -100,7 +100,7 @@
     </request>
   </interface>
 
-  <interface name="zwlr_layer_surface_v1" version="3">
+  <interface name="zwlr_layer_surface_v1" version="4">
     <description summary="layer metadata interface">
       An interface that may be implemented by a wl_surface, for surfaces that
       are designed to be rendered as a layer of a stacked desktop-like
@@ -241,6 +241,27 @@
           This setting is mainly intended for applications that need to ensure
           they receive all keyboard events, such as a lock screen or a password
           prompt.
+        </description>
+      </entry>
+      <entry name="on_demand" value="2" since="4">
+        <description summary="request regular keyboard focus semantics">
+          This requests the compositor to allow this surface to be focused and
+          unfocused by the user in an implementation-defined manner. The user
+          should be able to unfocus this surface even regardless of the layer
+          it is on.
+
+          Typically, the compositor will want to use its normal mechanism to
+          manage keyboard focus between layer shell surfaces with this setting
+          and regular toplevels on the desktop layer (e.g. click to focus).
+          Nevertheless, it is possible for a compositor to require a special
+          interaction to focus or unfocus layer shell surfaces (e.g. requiring
+          a click even if focus follows the mouse normally, or providing a
+          keybinding to switch focus between layers).
+
+          This setting is mainly intended for desktop shell components (e.g.
+          panels) that allow keyboard interaction. Using this option can allow
+          implementing a desktop shell that can be fully usable without the
+          mouse.
         </description>
       </entry>
     </enum>


dkondor:
wlroots branch for this (including updated layer-shell example):
https://github.com/dkondor/wlroots/tree/keyboard_interactivity_options2

Wayfire branch:
https://github.com/dkondor/wayfire/tree/layer_shell_keyboard

gtk-layer-shell branch (including updated demo):
https://github.com/dkondor/gtk-layer-shell/tree/new_keyboard_interactivity

example: cairo-dock branch using this to allow shift + click to start a new instance of a running program:
https://github.com/dkondor/cairo-dock-core/tree/layer_shell_keyboard_interactivity
(needs the updated gtk-layer-shell as well)

ammen99 (commented):
A few small nitpicks. Aside from that, LGTM, and I think this is a necessary addition if we want to grow layer-shell into the equivalent of `_NET_WM_STATE_ABOVE` and similar.

unstable/wlr-layer-shell-unstable-v1.xml
```diff
@@ -203,21 +204,85 @@
...
+
+          This is the default value, set for newly created layer shell surfaces.
+
+          This is useful for e.g. desktop widgets that display information or
+          only have simple interaction that can be manged by the pointer.
+        </description>
+      </entry>
+      <entry name="exclusive" value="1">
+        <description summary="request exclusive keyboard focus">
+          Request exclusive keyboard focus if this surface is above the desktop.
```
dkondor, line 229:
```suggestion
          Request exclusive keyboard focus if this surface is above the shell surface layer.
```

unstable/wlr-layer-shell-unstable-v1.xml
```diff
@@ -203,21 +204,85 @@
...
+
+      <entry name="no_focus" value="0">
+        <description summary="no keyboard focus is possible">
+          This value indicates that this surface is not interested in keyboard
+          events and the compositor should never assign it the keyboard focus.
+
+          This is the default value, set for newly created layer shell surfaces.
+
+          This is useful for e.g. desktop widgets that display information or
+          only have simple interaction that can be manged by the pointer.
```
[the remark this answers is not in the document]
dkondor:
```suggestion
          only have interaction with non-keyboard input devices.
```

dkondor:
Hi,
I'm wondering if there is a change to get this merged (or similar functionality -- I'm happy to work on alternate solutions as well). Let me know if I should open separate PRs with this functionality for wlroots and Wayfire (based on the above branches) or if I should give a longer overview of my motivation of this (extending the discussion in #32 ).

ammen99:
@dkondor You should open PRs at least in wlroots, so that the implementation can be reviewed (this is the usual procedure for protocols, protocol is drafted, then implementation, and then both are reviewed at the same time to ensure theory matches practice)

dkondor:
> @dkondor You should open PRs at least in wlroots, so that the implementation can be reviewed (this is the usual procedure for protocols, protocol is drafted, then implementation, and then both are reviewed at the same time to ensure theory matches practice)

Ok, sure :)

unstable/wlr-layer-shell-unstable-v1.xml
```diff
@@ -203,21 +204,85 @@
...
 
         Layer surfaces receive pointer, touch, and tablet events normally. If
         you do not want to receive them, set the input region on your surface
         to an empty region.
 
-        Events is double-buffered, see wl_surface.commit.
+        Keyboard interactivity is double-buffered, see wl_surface.commit.
       </description>
-      <arg name="keyboard_interactivity" type="uint"/>
+      <arg name="keyboard_interactivity" type="uint" enum="zwlr_layer_shell_v1.zwlr_layer_surface_v1.keyboard_interactivity"/>
```
emersion, line 285: This can just be `enum="keyboard_interactivity"` I think.

unstable/wlr-layer-shell-unstable-v1.xml
```diff
@@ -203,21 +204,85 @@
...
+
+          This is useful for e.g. desktop widgets that display information or
+          only have interaction with non-keyboard input devices.
+        </description>
+      </entry>
+      <entry name="exclusive" value="1">
+        <description summary="request exclusive keyboard focus">
+          Request exclusive keyboard focus if this surface is above the shell surface layer.
+
+          For layers above the shell surface layers, the seat will always give
```
emersion, line 231: This sounds a little bit like this is writing down compositor policy. I'd like to make it clear that compositors can decide to implement a different policy, e.g. don't give focus to any keyboard-interactive layer surface when a lockscreen is active.

unstable/wlr-layer-shell-unstable-v1.xml
```diff
@@ -203,21 +204,85 @@
...
+          exclusive keyboard focus to the top-most layer which has keyboard
+          interactivity set to exclusive. If this layer contains multiple
+          surfaces with keyboard interactivity set to exclusive, the compositor
+          determines the one receiving keyboard events in an implementation-
+          defined manner. In this case, no guarantee is made when this surface
+          will receive keyboard focus (if ever).
+
+          For layers below the shell surface layer, the seat will use
+          normal focus semantics.
```
emersion, line 240: It doesn't seem like this is backwards-compatible. I think e.g. Sway never gives focus in this case.

ammen99: Then sway isn't compliant with the current version of the protocol ;)
https://github.com/swaywm/wlr-protocols/blob/master/unstable/wlr-layer-shell-unstable-v1.xml#L209

unstable/wlr-layer-shell-unstable-v1.xml
emersion, line 231: This is the first time this protocol refers to a concept of "shell surface layer", I think? I wouldn't want to force compositors to have a single "shell surface layer" (which contains the regular toplevel surfaces if I understand your intention?).

The existing paragraph about the `layer` enum just describes what a compositor _typically_ does. Maybe a compositor has an always-on-top setting for a toplevel and will put the toplevel between the top and the overlay layer?

emersion: Ah, it seems like the wording comes from the original description for `set_keyboard_interactivity`.

unstable/wlr-layer-shell-unstable-v1.xml
```diff
@@ -203,21 +204,85 @@
...
+    <enum name="keyboard_interactivity" since="4">
+      <description summary="types of keyboard interaction possible for a layer shell surface">
+        Types of keyboard interaction possible for layer shell surfaces. The
+        rationale for this is twofold: (1) some applications are not interested
+        in keyboard events and not allowing them to be focused can improve the
+        desktop experience; (2) some applications will want to take exclusive
+        keyboard focus.
+      </description>
+
+      <entry name="no_focus" value="0">
```
emersion, line 216: Bikeshedding time: maybe `none` is a better name?

unstable/wlr-layer-shell-unstable-v1.xml
```diff
@@ -203,21 +204,85 @@
...
+
+          For layers below the shell surface layer, the seat will use
+          normal focus semantics.
+
+          This setting is mainly intended for applications that need to ensure
+          they receive all keyboard events, such as a lock screen or a password
+          prompt.
+        </description>
+      </entry>
+      <entry name="user" value="2">
```
emersion, line 247: Bikeshedding time: maybe we can come up with a better name?

unstable/wlr-layer-shell-unstable-v1.xml
emersion: Indeed, you're correct. This makes the change in this PR look less scary, because it just allows top/overlay layers to get the same behaviour as bottom/background layers which have keyboard interactivity set to 1.

I'd still prefer to allow compositors to not give focus if they don't want to, though.

unstable/wlr-layer-shell-unstable-v1.xml
dkondor: Yes, for this option, I took the current wording :)

I agree there is no guarantee to e.g. prevent stealing the focus from a lock screen either in the current protocol or this PR. I feel this could be a separate issue though.

emersion: Yes, you're right. We can fix it separately.

dkondor: Maybe instead of "shell surface layer", it could say explicitly that this refers to the top and overlay layers?

unstable/wlr-layer-shell-unstable-v1.xml
dkondor: Yes, sure

unstable/wlr-layer-shell-unstable-v1.xml
dkondor: I agree, but so far I couldn't think of anything better :)

unstable/wlr-layer-shell-unstable-v1.xml
emersion: This sounds good to me.

unstable/wlr-layer-shell-unstable-v1.xml
ammen99: `regular`? `on_demand`? (just ideas)

unstable/wlr-layer-shell-unstable-v1.xml
dkondor: I wonder if there could be more explicit wording (1) specifying whether the compositor is allowed to put non-layer-shell surfaces in any of the layers in the "layer" enum; and (2) noting that compositors will have additional "layers" beside the ones listed in the layer enum, that have a well-defined z-order relative to them (so e.g. if a fullscreen toplevel is above the "top" layer, a layer shell surface in the top layer will not receive exclusive keyboard focus).

unstable/wlr-layer-shell-unstable-v1.xml
dkondor: I'm changing this to optional. With this change, clients that want normal focus can just use the next option.

unstable/wlr-layer-shell-unstable-v1.xml
dkondor: I like on_demand

emersion:
I've taken the liberty to split your changes into 4 logical commits, to avoid having a single mega-commit that does many different things at once.

I've also changed these things:

- Removed the `since` argument for the error. Protocol errors aren't versioned.
- Moved the `since` argument from the enum to the `on_demand` entry. The other entries can be used with previous versions of the protocol.

Are you okay with these changes?

dkondor:
Sure, it all seems good. Thanks!

emersion:
OK, I've pushed the clarifications and minor improvements to master so that this PR only contains the new `on_demand` enum value.

This PR LGTM, I'll have a look at the compositor and client implementations linked in https://github.com/swaywm/wlr-protocols/pull/100#issuecomment-734774390.

emersion:
Can you open pull requests for these various branches? Would make it easier to add comments.

dkondor:
wlroots PR: https://github.com/swaywm/wlroots/pull/2555
wayfire PR: https://github.com/WayfireWM/wayfire/pull/962 (merged originally, as it should not break things)

I've just opened a PR for gtk-layer-shell:
https://github.com/wmww/gtk-layer-shell/pull/105

wmww (approved):
LGTM

emersion (approved):
Thanks!

PR #103, opened 2021-05-16
layer-shell: output property changes precede configure events

mstoeckl:
See commit message for justification. One place where the new guarantee would be helpful is 
in being able to safely remove a few lines of swaybg ([main.c#L196-198](https://github.com/swaywm/swaybg/blob/1.0/main.c#L196-L198)), in which the program redraws itself on receiving an output scale change. In practice, this redrawing is not necessary, because the associated configure event always comes afterwards; but this is not guaranteed by the existing spec.

The new text matches the current behavior of kwin and sway when changing output scales.

### DIFF
--- a/unstable/wlr-layer-shell-unstable-v1.xml
+++ b/unstable/wlr-layer-shell-unstable-v1.xml
@@ -346,6 +346,12 @@
 
         If the width or height arguments are zero, it means the client should
         decide its own window dimension.
+
+        If this event is associated with a change to the properties of the
+        output, then it will be sent after the events announcing the property
+        changes have been sent. For example, if an output changes size, then
+        xdg_output.logical_size, ..., xdg_output.done would be sent before
+        zwlr_layer_surface_v1.configure .
       </description>
       <arg name="serial" type="uint"/>
       <arg name="width" type="uint"/>


unstable/wlr-layer-shell-unstable-v1.xml
```diff
@@ -346,6 +346,12 @@
 
         If the width or height arguments are zero, it means the client should
         decide its own window dimension.
+
+        If this event is associated with a change to the properties of the
+        output, then it will be sent after the events announcing the property
+        changes have been sent. For example, if an output changes size, then
+        xdg_output.logical_size, ..., xdg_output.done would be sent before
```
emersion, line 353: I'd rather not mention xdg-output here. layer-shell clients aren't supposed to configure themselves depending on the outputs logical size.

emersion:
Hm. I don't know if this is desirable. There is no guarantee that a configure event will come after an output scale change.

ammen99:
> Hm. I don't know if this is desirable. There is no guarantee that a configure event will come after an output scale change.

My thoughts as well, output scale doesn't have anything to do with configure events since they are in logical coordinates. For example, resolution changes from 1080p to 2160p and scale changes from 1 to 2, we don't really need to reconfigure the surface. Or maybe the client has requested a fixed size (say 100x100), that won't change when the scale changes, right?

mstoeckl:
> For example, resolution changes from 1080p to 2160p and scale changes from 1 to 2, we don't really need to reconfigure the surface. Or maybe the client has requested a fixed size (say 100x100), that won't change when the scale changes, right?

Agreed. I had been thinking of background-type clients only, where in practice the configure size is the output size; but there is more than just sway+swaybg.

PR #106, opened 2021-09-05
screencopy: Add clip regions

any1:
This adds clip regions to the screencopy protocol. See https://github.com/swaywm/wlroots/pull/3166


emersion:

https://gitlab.freedesktop.org/wlroots/wlr-protocols/-/merge_requests/106

